PR #7, opened 2019-08-02
gui.hSlider with callback_finished

markotoplak:
callback_finished is called when the slider value stopped changing for at least 500 ms or when the slider is released.

A supporting class DelayedNotification was added to gui.py.

I opted to solve this problems with timing proxies on existing signals, because an alternative that would use sliderReleased signal would not solve similar problems for the keyboard and mouse wheel (+ it would require reimplementing lots of events).

This was done for biolab/orange3#3962

### DIFF
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -13,7 +13,7 @@ from collections import defaultdict
 import pkg_resources
 
 from AnyQt import QtWidgets, QtCore, QtGui
-from AnyQt.QtCore import Qt, QEvent, pyqtSignal as Signal
+from AnyQt.QtCore import Qt, QEvent, QObject, QTimer, pyqtSignal as Signal
 from AnyQt.QtGui import QCursor, QColor
 from AnyQt.QtWidgets import (
     QApplication, QStyle, QSizePolicy, QWidget, QLabel, QGroupBox, QSlider,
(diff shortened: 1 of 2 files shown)


orangewidget/gui.py
```diff
@@ -1083,8 +1083,42 @@ def appendRadioButton(group, label, insertInto=None,
...
+        if self._did_notify and self._last_notified == self._last_value:
+            return
+        self._last_notified = self._last_value
+        self._did_notify = True
+        self.notification.emit()
+
+
 def hSlider(widget, master, value, box=None, minValue=0, maxValue=10, step=1,
-            callback=None, label=None, labelFormat=" %d", ticks=False,
+            callback=None, callback_finished=None, label=None, labelFormat=" %d", ticks=False,
```
lanzagar: Do you think there will be situations when a user will want to specify different functions for `callback` _and_ `callback_finished`?
Because otherwise it might be better to just keep using `callback` and introduce a new parameter like `delay` or something similar, where you can specify e.g. 500 (ms). If it is None you get the old functionality, and if you provide a delay the callback is used like callback_finished in your new version.
(The `delay` parameter could also be just `True`/`False` in you don't think being able to choose a different length is useful)

markotoplak: Some sliders show chosen values in label, like the slider in the Logistic Regression widget. Especially there, where it is non-linear, it is good to see what we are choosing. I could not think of any other way that to have two callbacks.

janezd: I agree with @markotoplak, sliders often update some label instantly, but should update the widget when the action is finished.

But can't we achieve this by connecting `sliderMoved` to `callback` and slider released to `callback_finished`? Do we need this delay?

markotoplak: @janezd, if we used sliderReleased as you suggested we would have to handle keyboard presses (at least on my system I can focus the slider and control it with a keyboard) and mouse wheel events. The current implementation calls callback_finished on sliderReleased without the delay.

lanzagar: Yes, in that case two callbacks seem fine.
The delay itself would probably not be worth it as it is useful only in rare specific circumstances. But if it helps to simplify the implementation, I guess we get a bonus feature alongside not having to worry about many different event types.

PR #19, opened 2019-10-08
Fix incorrectly set machine id

irgolic:
##### Description of changes
In certain cases, the UUID python object was written to config as `machine_id` instead of a string. This change casts it to string before writing it to config.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/workflow/errorreporting.py
+++ b/orangewidget/workflow/errorreporting.py
@@ -231,7 +231,7 @@ class ErrorReporting(QDialog):
         if settings.contains('error-reporting/machine-id'):
             machine_id = settings.value('error-reporting/machine-id')
         else:
-            machine_id = uuid.uuid4()
+            machine_id = str(uuid.uuid4())
             settings.setValue('error-reporting/machine-id', machine_id)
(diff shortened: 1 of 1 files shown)


lanzagar:
Did this cause any problems?
This change will cast it to string for _newly generated_ ids from now on. Will the already stored non-string ids still cause any issues?

irgolic:
Yes, here's an example of the error upon attempted retrieval of `machine-id` (thanks @thocevar):
```
ERROR:__main__:Failed to pull notification feed
Traceback (most recent call last):
 File "C:\Orange\orange3\Orange\canvas\__main__.py", line 236, in run
   'User-Agent': ua_string(),
 File "C:\Orange\orange3\Orange\canvas\__main__.py", line 69, in ua_string
   machine_id = settings.value("error-reporting/machine-id", "", str)
TypeError: unable to convert a QVariant of type 1024 to a QMetaType of type 10
```

The faulty code was commited on September 13th, **after** the most recent release, so it should be ok.

PR #22, opened 2019-11-02
Break up module gui

janezd:
Module gui is one huge mess with a long history of well-intended sins. The least we can do is to separate code for components (like checkBox, comboBox) into separate files, together with their callbacks, callfronts etc.

@ales-erjavec, do you think this would be a good idea and have I approached it correctly so far? (See https://github.com/biolab/orange-widget-base/commit/106d32cba73d71cd90ff1fa9b359c9504beb1a73, not the two commits coming from #21.)

Next, I plan to clean up individual parts. For instance, most functions that construct controls resemble constructors, except that they store data as extra attributes in QObjects that they construct in return. Thus, functions like `lineEdit` etc. should probably be classes. These classes would then include various callback functions and other auxililiary stuff as methods.

I have left some code in `__init__`. I would move that out, too.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


ales-erjavec:
> do you think this would be a good idea and have I approached it correctly so far? (See 106d32c, not the two commits coming from #21.)

Beats me. I don't see how that would help. I would just advise against using `gui` and be done with it.

PR #27, opened 2019-11-15
Context: Add __eq__ and __hash__

janezd:
##### Issue

Fixes https://github.com/biolab/orange3/issues/4185.

As @ales-erjavec explained, Orange always asks whether to save the schema (at least) if if contains widgets with context settings. This happens because context settings are never equal.

##### Description of changes

This PR provides `Contexts.__eq__` (and, for consistency, also `Context.__hash__`) that ignore time stamps. This should solve a general problem, but individual widgets may still have hidden modified values.

##### Includes
- [X] Code changes

### DIFF
--- a/orangewidget/settings.py
+++ b/orangewidget/settings.py
@@ -693,7 +693,7 @@ class ContextSetting(Setting):
 
 
 class Context:
-    """Class for data thad defines context and
+    """Class for data that defines context and
     values that should be applied to widget if given context
     is encountered."""
     def __init__(self, **argkw):
@@ -707,6 +707,13 @@ class Context:
             if nc in state:
                 del state[nc]
         return state
+
+    def __eq__(self, other):
+        self_state = self.__getstate__()
+        del self_state["time"]
+        other_state = other.__getstate__()
+        del other_state["time"]
+        return self_state == other_state
 
 
 class ContextHandler(SettingsHandler):
--- a/orangewidget/settings.py
+++ b/orangewidget/settings.py
@@ -701,17 +701,10 @@ class Context:
         self.values = {}
         self.__dict__.update(argkw)
 
-    def __getstate__(self):
-        state = dict(self.__dict__)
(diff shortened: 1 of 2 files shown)


orangewidget/settings.py
```diff
@@ -708,6 +708,17 @@ def __getstate__(self):
...
+    def _stable_state(self):
+        state = self.__getstate__()
+        del state["time"]
+        return state
+
+    def __eq__(self, other):
+        return self._stable_state() == other._stable_state()
+
+    def __hash__(self):
+        return hash(list(self.__getstate__.items()))
```
astaric: this does not work. __getstate__ is a function, you probably need to call it. list is not hashable, and neither are dictionaries (self.values). self.values can contain anything, not just hashable values.

Does context really need __hash__? Is it ever used in sets/keys of a dict?

janezd: Please let us attribute this to late hour. The lesson is to always write a test.

I wrote `__hash__` because I thought I have to provide one, even if just to raise an exception (but then I thought I'd just implement a "real one"). I didn't know (or forgot) that defining `__eq__` already makes a class unhashable unless `__hash__` is defined explicitly.

janezd:
I added another commit: I removed `__getstate__`, which existed only to remove the context values in the `no_copy` list. To my search,`no_copy` is never used in our code, it is not documented and I very much doubt anybody outside discovered it. Using it would probably be a sign of bad design.

@astaric?

astaric:
noCopy was copied from Orange2
https://github.com/biolab/orange3/commit/cae097749e098191b6789e0bc4ca68b85851c749
, and later renamed to no_copy.

The only usage I could find was here:
https://github.com/biolab/orange3/commit/bb4404a5358534cd629d5851c26b0caed9eeaf13#diff-2281ecd412ed7edd749aca4486d1f338L398
but has since been removed. Should be ok to remove.

PR #33, opened 2019-12-27
Add hide/show indicator in splitter between control and main area

janezd:
##### Issue

Windows users see no indicator in splitter. On other systems, the indicator does not suggest that clicking it will show/hide the control area.

##### Description of changes

Make the splitter wider and add an arrow that changes directions on show/hide.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/widget.py
+++ b/orangewidget/widget.py
@@ -16,7 +16,7 @@ from AnyQt.QtWidgets import (
 from AnyQt.QtCore import (
     Qt, QObject, QEvent, QRect, QMargins, QByteArray, QDataStream, QBuffer,
     QSettings, QUrl, QThread, pyqtSignal as Signal, QSize)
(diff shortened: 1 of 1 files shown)


orangewidget/widget.py
```diff
@@ -330,12 +334,40 @@ def sizeHint(self):
...
+
+            def paintEvent(self, event):
+                super(QSplitterHandle, self).paintEvent(event)
+                painter = QPainter(self)
+                painter.setPen(QColor(160, 160, 160))
+                w = self.width() - 2
+                if self.controlAreaShown:
+                    x0, x1 = 1, w - 1
+                else:
+                    x0, x1 = w - 1, 1
```
lanzagar: Starting 2 pixels away from the left border looks better for me than 1.
+ as @janezd noticed, the above code does not center (1-3 margins, instead of 2-2)

PR #35, opened 2020-01-07
WidgetTest: Check types of settings

### LINKED ISSUE #20: Check types of settings
As @ales-erjavec wrote in #15, settings can only be lists, tuples, dicts, ints, floats, strings, bytes and None. Violations cause problems that may be difficult to diagnose.

This cannot be checked in unittests or in widget's meta class. A frequently called function would need to have an `assert` statement for checking types.

VesnaT:
##### Issue

Fixes #20

##### Description of changes
Add settings checking into `tearDown` of `WidgetTest`.
The change causes the following tests (in biolab/orange3) to fail:

- [ ] test_owcolor (saves`Orange.widgets.data.owcolor.AttrDesc`)
- [ ] test_oweditdomain (references OWColor)
- [x] test_owfeaturestatistics (saves np.ndarray, fixed in https://github.com/biolab/orange3/pull/4375)
- [ ] test_owpivot (saves enum)
- [ ] test_owpreprocess (saves enum)
- [ ] test_owpythonscript (saves `Orange.widgets.data.owpythonscript.Script`)
- [x] test_owrank (fixed by https://github.com/biolab/orange3/pull/4315)
- [ ] test_report (saves enum)
- [ ] test_owdistancemap (saves range)
- [ ] test_owboxplot (saves `Orange.data.filter.Values`)
- [ ] test_owfile (saves `RecentPath`, but it does not fail, because parent `tearDown` is not invoked)

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/settings.py
+++ b/orangewidget/settings.py
@@ -41,6 +41,8 @@ import warnings
 from operator import itemgetter
 from typing import Any, Optional, Tuple
 
+import numpy as np
+
 from orangewidget.gui import OWComponent
 
 log = logging.getLogger(__name__)
@@ -48,7 +50,8 @@ log = logging.getLogger(__name__)
 __all__ = [
     "Setting", "SettingsHandler", "SettingProvider",
     "ContextSetting", "Context", "ContextHandler", "IncompatibleContext",
-    "SettingsPrinter", "rename_setting", "widget_settings_dir"
+    "SettingsPrinter", "rename_setting", "widget_settings_dir",
+    "is_setting_type_supported"
 ]
 
 _IMMUTABLES = (str, int, bytes, bool, float, tuple)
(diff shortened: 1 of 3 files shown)


janezd:
What if this PR showed warnings instead of raising errors for now? It would warn that after 2/1/21 they would change into errors, though.

In its current form, we shouldn't merge it. :)

janezd:
Why is OWFile not on this list although it stores a list of `RecentPath`?

@ales-erjavec, did I understand correctly that the idea is to avoid pickling settings?

Context settings are stored as instances of `ContextSetting`. `OWFile` stores `RecentPath` instances. `OWColor` stores (essentially) namedtuples with some properties... Adapting this would burden the widgets with code that would essentially perform serialization of settings. I wonder whether the benefits of having json outweigh the trouble and the extra code for conversion (with its maintenance and potential bugs).

Or are there other reasons for limiting the allowed types for settings?

ales-erjavec:
>>> did I understand correctly that the idea is to avoid pickling settings?

Yes.

VesnaT:
>>> Why is OWFile not on this list although it stores a list of RecentPath?

Because `TestOWFile` overrides `tearDown()`.

janezd:
Superseeded by #62.

PR #36, opened 2020-01-08
[ENH] Refactor statusbar, prettify

irgolic:
(edited 11th Feb)
##### Description of changes
- Played around with content margins
- Elide warning/error message
- Fix in/out message size
- Remove in/out message size grip
- In/out messages always have a popup
- Remove first line from popup/tooltip

Before:

statusbar-before

After:

statusbar-after

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/utils/messagewidget.py
+++ b/orangewidget/utils/messagewidget.py
@@ -293,6 +293,50 @@ def summarize(messages):
                    textFormat=Qt.RichText)
 
 
+class ElidingLabel(QLabel):
+    def __init__(self, elide=False, **kwargs):
+        super().__init__(**kwargs)
+        self.__elide = elide
+        self.__originalText = ""
+
+    def resizeEvent(self, event):
+        if self.__elide:
+            self.__setElidedText(self.__originalText)
+
+    def __setElidedText(self, text):
+        fm = self.fontMetrics()
+
+        # Qt sometimes elides even when text width == target width
+        width = self.width() + 1
+
+        elided = fm.elidedText(text, Qt.ElideRight, width)
+        super().setText(elided)
+
+    def setText(self, text):
+        self.__originalText = text
+        if self.__elide:
+            self.__setElidedText(text)
+        else:
+            super().setText(text)
+
+    def sizeHint(self):
+        fm = self.fontMetrics()
+        w = fm.width(self.__originalText)
+        h = super().minimumSizeHint().height()
+
+        return QSize(w, h)
+
+    def setElide(self, enabled):
+        if self.__elide == enabled:
+            return
+
+        self.__elide = enabled
+        if enabled:
+            self.__setElidedText(self.__originalText)
+        else:
+            super().setText(self.__originalText)
+
+
 class MessagesWidget(QWidget):
     """
     An iconified multiple message display area.
@@ -318,7 +362,7 @@ class MessagesWidget(QWidget):
 
     Message = Message
 
-    def __init__(self, parent=None, openExternalLinks=False,
+    def __init__(self, parent=None, openExternalLinks=False, elideText=False,
                  defaultStyleSheet="", **kwargs):
         kwargs.setdefault(
             "sizePolicy",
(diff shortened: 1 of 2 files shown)


orangewidget/utils/messagewidget.py
```diff
@@ -541,6 +541,8 @@ def __update(self):
             fulltext = ""
         self.__fulltext = fulltext
         self.setToolTip(self.__styled(self.__defaultStyleSheet, fulltext))
+        self.setVisible(True)
```
ales-erjavec: Restore the visible state as it was at entry (use `not isHidden()` to get the correct value to restore).

ales-erjavec: Still, this seems wrong.

I think `self.layout().activate()` below is the confounding factor in the jittering update.
I would suggest:
1) removing that line
2) left align the icon widget so it preserves its absolute position within MessageWidget event when it's wider then necessary.

```diff
diff --git a/orangewidget/utils/messagewidget.py b/orangewidget/utils/messagewidget.py
index 4ea6f2cf8..7b3f624eb 100644
--- a/orangewidget/utils/messagewidget.py
+++ b/orangewidget/utils/messagewidget.py
@@ -356,7 +356,7 @@ class MessagesWidget(QWidget):
         self.setLayout(QHBoxLayout())
         self.layout().setContentsMargins(2, 1, 2, 1)
         self.layout().setSpacing(0)
-        self.layout().addWidget(self.__iconwidget)
+        self.layout().addWidget(self.__iconwidget, alignment=Qt.AlignLeft)
         self.layout().addSpacing(4)
         self.layout().addWidget(self.__textlabel)
         self.layout().addWidget(self.__popupicon)
@@ -552,7 +552,6 @@ class MessagesWidget(QWidget):
         else:
             self.__popuptext = fulltext
         self.__popupicon.setVisible(bool(self.__popuptext))
-        self.layout().activate()
 
     def mousePressEvent(self, event):
         if event.button() == Qt.LeftButton:
```

irgolic: What role does `self.layout().activate()` play here? Left-align solves the jittering with or without `.activate()`. I'll throw it out anyway, it's probably called implicitly at some point.

thocevar:
Before and after on Win 10.

image

ajdapretnar:
Would bigger spaces lead to problems with smaller widgets? Say Distances + some Warning?

markotoplak:
Before and after on linux with dwm.

before
after

orangewidget/utils/messagewidget.py
irgolic: Scratch that, I'm leaving it in, else this might happen: 
Screenshot 2020-01-14 at 14 14 44

markotoplak:
Current version on linux with dwm:
newest

irgolic:
Thanks @thocevar, @markotoplak! Looks good to me.

@ajdapretnar This happens. I'd assume this isn't the only instance of a warning being cut off, even without the extra margins. I'm thinking, let's display the warning message if there's space in the status bar, otherwise just display the icon, storing the message in its tooltip.
Screenshot 2020-01-14 at 14 40 15

ajdapretnar:
> let's display the warning message if there's space in the status bar, otherwise just display the icon, storing the message in its tooltip

Agree.

irgolic:
@ajdapretnar I've opted to elide the text rather than hide it completely.

In/out messages are now fixed in size. I've removed the popup icon – it looks like a grip rather than a hoverable/clickable item. The messages already change their background on, making it obviously interactable. They now also show a popup on no input/output. 

Before:

statusbar-before

After:

statusbar-after

Before merging someone should check if the status bar and its tooltips/popups look drastically different on Windows/Linux (it'll probably be fine).

markotoplak:
Looks good on Linux.

orangewidget/widget.py
```diff
@@ -638,21 +662,28 @@ def __info(self):
...
+                in_out_msg = sb.findChild(QWidget, "in-out-msg")
+
                 # Insert a separator if these are not the first elements
-                # TODO: This needs a better check.
-                if sb.findChildren(SimpleButton):
-                    sb.addWidget(QFrame(frameShape=QFrame.VLine))
-                sb.addWidget(in_msg)
-                sb.addWidget(out_msg)
+                buttons = sb.findChild(QWidget, "buttons")
+                if buttons.layout().count() != 0:
```
ales-erjavec: `buttons` can be None here if `want_main_layout` is False (the creation of the `buttons` should probably be moved to `statusBar` though).

PR #46, opened 2020-02-24
Added search option to comboBox

PrimozGodec:
##### Issue

ComboBoxSearch implemented but not included in comboBox fuction.
ComboBoxSearch does not implement maximumContentsLenght, which is implemented in ComboBox.

##### Description of changes
- Adding a searchable attribute to comboBox function - when True ComboBoxSearch used instead of ComboBox
- Deprecating maximumContentsLenght for ComboBox - reason not really used in practice 
- Setting maximumContentsLenght to the default value of 25.
- Implementing the same behavior for ComboBoxSearch (default max width 25).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -23,7 +23,9 @@ from AnyQt.QtWidgets import (
 
 from orangewidget.utils import getdeepattr
 from orangewidget.utils.buttons import VariableTextPushButton
-from orangewidget.utils.combobox import ComboBox as OrangeComboBox
+from orangewidget.utils.combobox import (
+    ComboBox as OrangeComboBox, ComboBoxSearch as OrangeComboBoxSearch
+)
 from orangewidget.utils.itemmodels import PyListModel
 
 __re_label = re.compile(r"(^|[^%])%\((?P<value>[a-zA-Z]\w*)\)")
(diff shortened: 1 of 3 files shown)


orangewidget/gui.py
```diff
@@ -1458,9 +1461,12 @@ def comboBox(widget, master, value, box=None, label=None, labelWidth=None,
     else:
         hb = widget
 
-    combo = OrangeComboBox(
-        hb, maximumContentsLength=maximumContentsLength,
-        editable=editable)
+    if search:
+        combo = OrangeComboBoxSearch(hb)
```
lanzagar: Does this mean that setting `seach=True` disables/ignores the `maximumContentsLength` and `editable` parameters? How do we feel about that?
Also this interaction is not mentioned anywhere!

orangewidget/gui.py
```diff
@@ -24,6 +24,7 @@
 from orangewidget.utils import getdeepattr
 from orangewidget.utils.buttons import VariableTextPushButton
 from orangewidget.utils.combobox import ComboBox as OrangeComboBox
+from orangewidget.utils.combobox import ComboBoxSearch as OrangeComboBoxSearch
```
lanzagar: these two imports are from the same module and could be joined

orangewidget/utils/combobox.py
```diff
@@ -13,6 +14,10 @@
 )
 
 
+# we want to have combo box maximally 25 characters wide
+MAXIMUM_CONTENTS_LENGTH = 25
```
lanzagar: I would still leave this as a class constant/attribute so that it is at least possible to change it per instance in rare cases when that is needed (although setting it in the constructor is being deprecated).

PrimozGodec: I put back the class attributes but still left this constant here - now attributes from both ComboBox, ComboBoxSearch get the default value from this constant. I also added setter for maximuContentLength to ComboBoxSearch that they work the same.

PR #48, opened 2020-02-26
[FIX] SvgFormat: Set 'viewBox' instead of 'size' to initiate the svg viewport

ales-erjavec:
##### Issue

QSvgGenerator emits width and height in mm units, while the exported
geometry is unitless. 'viewBox' is emitted unitless and matches the
geometry.

##### Description of changes

Use `QSvgGenerator.viewBox` instead of `QSvgGenerator.size` to set exported svg viewport

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


ales-erjavec:
Screen shots from Inkscape:

* before

Screenshot 2020-02-26 at 15 17 05

* after
Screenshot 2020-02-26 at 15 19 20

PR #53, opened 2020-03-24
[FIX] open_filename_dialog: Fix empty filter returned from QFileDialog.getOpenFileName

janezd:
##### Issue

https://github.com/biolab/orange3/issues/4537: `getOpenFileName` sometimes returns an empty filter. We can't reproduce the problem, so this is a blind fix, with a test.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/orangewidget/utils/filedialogs.py
+++ b/orangewidget/utils/filedialogs.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import typing
+from fnmatch import fnmatch
 from typing import Tuple
 
 from AnyQt.QtCore import QFileInfo, Qt
@@ -121,7 +122,7 @@ def open_filename_dialog_save(start_dir, start_filter, file_formats):
         return filename, format, filter
 
 
-def open_filename_dialog(start_dir, start_filter, file_formats,
+def open_filename_dialog(start_dir: str, start_filter: str, file_formats,
                          add_all=True, title="Open...", dialog=None):
     """
     Open file dialog with file formats.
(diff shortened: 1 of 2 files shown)


janezd:
@robertcv commented we should discover why it happens instead of applying patches. Check Qt code -- when does it return an empty filter?!

Doesn't work for *.tab.gz.

Hrovatin:
I have tried this on latest Orange and widget base masters. I get an error without this PR but no error when I use this PR.

janezd:
> I have tried this on the latest Orange and widget base masters. I get an error without this PR, but no error when I use this PR.

Good to hear -- thanks for checking!

Let's still keep it open until I also do what @robertcv suggested.

janezd:
> @robertcv commented we should discover why it happens instead of applying patches. Check Qt code -- when does it return an empty filter?!

The filter is retrieved here: https://github.com/qt/qtbase/blob/dev/src/widgets/dialogs/qfiledialog.cpp#L1531. On Linux, Qt doesn't use the native dialog but its own, so it would appear that the `currentText` in the combo is empty.

> Doesn't work for *.tab.gz.
> 
> This function returns a filter and the caller does it again (for the "all" filter). This could be implemented in a single place, not two.

I have now replaced this PR's previous code with an adaptation from `FileReader.get_reader`, which is called for unknown filters, including the "all" filter. The `OWFile._get_reader` still contains a call to `get_reader`, which is now probably redundant and won't happen, but I daren't remove it.

@Hrovatin, this fix is essentially the same, but with different code. Can I ask you to try it again?

Hrovatin:
It works.

markotoplak:
@janezd, this PR broke using the function with `dialog=QFileDialog.getOpenFileNames`, which I use in OWMultiFile (quasars/orange-spectroscopy#431). That dialog returns a list of files. First, I though of just making my own function and get over with it, but then I reviewed this PR and think that it should be mostly reverted.

This function is now trying to be too smart. I like the possibility that the caller knows that the user did not select any specific filters. If additional guessing was needed, that could be done outside - and already is done in get_readers. Plus, it duplicates get_readers while also doing it wrong: it ignores reader priorities and just uses formats in the order that was given (well, maybe not wrong, but the behaviour is changed :) ).

I would like to remove this PR. Wouldn't it be enough to fix the bug Karin was having if only the All formats filter was chosen if the output filter was not one of the listed ones? Because then the reader is chosen by get_readers anyway. What do you think, @janezd?

markotoplak:
I opened a PR with my suggestion in #64.

PR #54, opened 2020-03-26
Searchable combo filter: Move popup below focus frame

janezd:
On (my) mac, the popup covers the border around the line edit, making it i-can't-watch-this ugly.

Screenshot 2020-03-26 at 12 56 27

I think it should be shown a few pixels lower.

Screenshot 2020-03-26 at 13 06 31

How does this look on Linux and Windows?

A perhaps better alternative could be to not show the frame at all.

### DIFF
--- a/orangewidget/utils/combobox.py
+++ b/orangewidget/utils/combobox.py
@@ -1,4 +1,4 @@
-import warnings
+import sys
 from typing import Optional
 
 from AnyQt.QtCore import (
@@ -238,6 +238,10 @@ class ComboBoxSearch(QComboBox):
         popuprect_origin = style.subControlRect(
             QStyle.CC_ComboBox, opt, QStyle.SC_ComboBoxListBoxPopup, self
         )  # type: QRect
+        if sys.platform == "darwin":
+            slmargin = self.__searchline.style() \
+                .pixelMetric(QStyle.PM_FocusFrameVMargin)
+            popuprect_origin.adjust(slmargin / 2, 0, -slmargin * 1.5, slmargin)
         popuprect_origin = QRect(
             self.mapToGlobal(popuprect_origin.topLeft()),
             popuprect_origin.size()


orangewidget/utils/combobox.py
```diff
@@ -244,6 +244,8 @@ def showPopup(self):
         popuprect_origin = style.subControlRect(
             QStyle.CC_ComboBox, opt, QStyle.SC_ComboBoxListBoxPopup, self
         )  # type: QRect
+        popuprect_origin.moveTop(
+            self.__searchline.style().pixelMetric(QStyle.PM_FocusFrameVMargin))
```
janezd: *Doubling* this distance has it's charm, too. Try it.

thocevar:
Windows (current, single, double)

image

ajdapretnar:
> the popup covers the border around the line edit, making it i-can't-watch-this ugly

Really? This bothers you and not the fact that the dropdown is not aligned with the line edit field??? 😆

lanzagar:
It is very similar on Linux as on Win:
(current, single, double)
current
single
double

I like the first two, but find the last one too disconnected personally.
So since 1) is not nice on mac, I guess I vote for number 2 (single space)

ajdapretnar:
I prefer 0 on Win and Mac. Is it not possible to do sth with the border on Mac? Make it thinner?
Otherwise 1 is better than 2.

janezd:
I'm sad to hear that nobody likes the last option. :)

It's weird that the second option (single) is disconnected on Linux and Windows. I believed that the metric I used gives the width of that frame.

I agree that it would be by far that best to eliminate the wide frame on Mac and keep popup as it is. If anybody knows a simple way to do it -- welcome.

irgolic:
A screenshot of option 2 on Mac:
Screenshot 2020-03-27 at 16 45 50

Current looks fine on Windows and Linux, Mac could use some makeup though.
It seems as if Windows and Linux don't even use a focus frame, which is why the dropdown looks disconnected when the margin is added.

I like the single contiguous control (option 1) better than the detached alternative (option 2).

When implementing this, be careful to cover the flipped dropdown edge case too.
Screenshot 2020-03-27 at 16 39 38

It's strange how the dropdown is of correct width/origin point on Win/Linux, but not on Mac. Perhaps it's related to there being a focus margin.

Or maybe, `SC_ComboBoxListBoxPopup` shouldn't be used as a size reference.
Screenshot 2020-03-27 at 16 49 15

janezd:
`ComboBoxSearch.__init__` includes the following:

```
        self.__searchline.setAttribute(Qt.WA_MacShowFocusRect, False)
        self.__searchline.setFocusProxy(self)
```

The first line currently has no effect.

I do not understand the purpose of the second line. Why would an object be a focus proxy of itself? If I remove the second line, the first line removes the focus frame. @ales-erjavec?

ales-erjavec:
The first line removes the mac focus rect from the contained line edit component.

The second sets the parent combo box as the line edit's focus proxy, effectively moving the focus to the combo box (it does not set itself as a proxy).

janezd:
I overlooked it's a different `self`. :) Sorry. But what breaks if we remove it?

The first line doesn't work (at least for me), the blue frame is still present despite the flag. If I remove the second line, the frame disappears.

ales-erjavec:
The fist line is removing the frame within

Screen Shot 2020-04-06 at 10 27 42

Although I need to remove both lines to display this (which probably means the first line is superfluous; since the focus is moved anway).

It has no effect on the focus frame around the whole combo box. In fact the second (setFocusProxy) line is intended to have the focus box on the whole of the combo. The focus has to have a visual indicator and the combo box is the best (most obvious) place.

janezd:
Can we remove `setFocusProxy`? It looks as desired, and I haven't found any change in behaviour -- but maybe I'm not trying the right thing.

orangewidget/utils/combobox.py
```diff
@@ -238,6 +238,8 @@ def showPopup(self):
         popuprect_origin = style.subControlRect(
             QStyle.CC_ComboBox, opt, QStyle.SC_ComboBoxListBoxPopup, self
         )  # type: QRect
+        sl_margin = self.__searchline.style().pixelMetric(QStyle.PM_FocusFrameVMargin)
+        popuprect_origin.adjust(sl_margin/2 - 1, 0, -sl_margin*1.5 + 1, sl_margin)
```
janezd, line 242: @irgolic, I get a perfect alignment if I remove `- 1` and `+ 1`. Could you check?

It would be nice to know where 1.5 comes from. It may be some other margin we need to take into account...

irgolic: You're right, I get perfect alignment without the `+/- 1` too. Go ahead and remove it.

No idea where the 1.5 comes in, I'm not sure why the box is so long without adjustments.

PR #62, opened 2020-05-01
Typed and json-like-storable Settings

janezd:
##### Issue

Implements #20.

##### Description of changes

PR defines a set of allowed types and requires proper annotations when the type is not deducible from the default.

- The list of allowed types includes `IntEnum` and `NamedTuple`, as well as `set`, `tuple` and `bytes` which can easily be converted before encoding to json.
- Context handlers can add other valid types, though they are expected to encode them, like, for instance, `Variable` which is already encoded into tuple.
- If setting's default is an empty `list`, `set`, `dict` or `tuple`, it must be annotated, e.g. as `List[int]`. All these must be homogenous; named `tuple` is an exception.
- Type checking is deep, e.g. a `List` must contain only elements of allowed types. Dict's keys must be `int`, `float`, `str` or `bool`, to match those allowed in json.

Also:
- `savedWidgetGeometry` is base64-encoded into strings (this may be reverted because `bytes` will be allowed).

**To do**: Handlers will provide encoding to json and decoding from it. `IntEnum` will become an int, `bytes` will be base64-encoded, and `NamedTuple`, `set`, `tuple` will be encoded as lists. Decoding to proper type can be done because types are annotated. When encoding to json fails, handler will fallback to pickle.

As things stand now, types like `Variable` are already encoded to tuple when settings are packed into a dict. Packed dicts thus contain only valid types that are described above. Hence, encoding will be done in the base `SettingsHandler`.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/settings.py
+++ b/orangewidget/settings.py
@@ -7,20 +7,21 @@ marked as settings are read from disk. When schema is loaded, attribute values
 are set to one stored in schema.
 
 Allowed setting types are
-- str
-- bool
-- int (Integral values are converted to int when saving and kept int at load)
-- float (Number values are converted to float when saving and kept int at load)
-- bytes (implicitly b64encoded/decoded),
+- int, float, bool, str, bytes,
 - IntEnum (implicitly encoded/decoded as int),
 
 and the following generics, whose elements must be one of allowed types
-- Optional
-- List
-- Dict (keys must not be generics; values can be any allowed type)
-- Set (converted to list and back)
+- List, Dict, Tuple,
 - NamedTuple (implicitly converted to tuple and back)
-- Tuple with fixed number of elements with annotated types
+- Set (converted to list and back)
+- Optional
+- Union of types that can be saved as literals.
+
+Types that can be saved as literals are
+- int, float, bool, str, bytes
+- list, tuple of types that can be saved as literals
+- dicts whose keys and values can be saved as literals.
+
 
 Derived setting handlers may add additional types if the provide proper
 conversion.
@@ -768,24 +769,20 @@ class SettingsHandler:
                 args = getattr(tp, "__annotations__", None)
                 return args is not None \
                        and all(map(cls.is_allowed_type, args.values()))
+            for type_handler in cls.type_support:
+                if type_handler.supports_type(tp):
+                    return True
 
         orig, args = get_origin(tp), get_args(tp)
-        if orig is None:
-            return False
         if orig is Union:
             return len(args) == 2 and type(None) in args \
                    and cls.is_allowed_type(cls._non_none(args))
         if orig in (list, set) \
                 or orig is tuple and len(args) == 2 and args[1] is ...:
             return cls.is_allowed_type(args[0])
-        if orig is tuple:
+        if orig in (tuple, dict):
             return all(map(cls.is_allowed_type, args))
-        if orig is dict:
-            # json.dumps will convert int and float into str, but we decode
-            # them back. We could allow bool, but it would complicate decoding
-            # and is not of much use anyway
-            return args[0] in (str, int, float) \
-                   and cls.is_allowed_type(args[1])
+        return False
 
     @classmethod
     def check_type(cls, value, tp) -> bool:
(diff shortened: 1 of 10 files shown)


lanzagar:
Python Script does not work for me using this PR:
Latest master of Orange and canvas, cleared settings. After opening Orange and adding the Python Script widget I immediately get an error:
```
...
  File "/home/lan/dev/orange3/Orange/widgets/data/owpythonscript.py", line 399, in <module>
    class OWPythonScript(OWWidget):
  File "/home/lan/dev/orange-widget-base/orangewidget/widget.py", line 73, in __new__
    SettingsHandler.create(cls, template=cls.settingsHandler)
  File "/home/lan/dev/orange-widget-base/orangewidget/settings.py", line 497, in create
    setting_handler.bind(widget_class)
  File "/home/lan/dev/orange-widget-base/orangewidget/settings.py", line 508, in bind
    self.provider = SettingProvider(widget_class)
  File "/home/lan/dev/orange-widget-base/orangewidget/settings.py", line 250, in __init__
    type_hints = get_type_hints(provider_class)
  File "/usr/lib/python3.8/typing.py", line 1232, in get_type_hints
    value = _eval_type(value, base_globals, localns)
  File "/usr/lib/python3.8/typing.py", line 270, in _eval_type
    return t._evaluate(globalns, localns)
  File "/usr/lib/python3.8/typing.py", line 518, in _evaluate
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
NameError: name '_ScriptData' is not defined
```

janezd:
@ales-erjavec, is there any reason to define `_ScriptData` only when `TYPE_CHECKING`?

@lanzagar, even so, this PR shouldn't break any code. I'll fix it.

ales-erjavec:
> is there any reason to define _ScriptData only when TYPE_CHECKING?

To avoid adding typing_extensions to run dependencies. _ScriptData is TypedDict, which is new in Python 3.8

janezd:
@lanzagar, this was so long ago that I forgot whether I've fixed the problem with PythonScript. I can no longer replicate it. Could you please check?

lanzagar:
Python Script now works for me as well. Looks like it was fixed :)

janezd:
Note to self: it may make sense to use a deeper inspection. Current code allows lists of anything. We could require annotating `Setting([])` (and `list` would not be a valid annotation).

Annotations like `List[Tuple[int, int, str]]` would be deep-checked, and we could also implement proper conversion for tuples (that are currently json-decoded to lists, which may cause problems).

janezd:
> Python Script now works for me as well. Looks like it was fixed :)

It was.

Some more propaganda: yesterday I was debugging settings for Mosaic. Checking out this PR made it much simpler, because I could read the saved .ows file.

orangewidget/settings.py
```diff
@@ -381,63 +381,58 @@ def store_initialization_data(self, initialization_data: dict) -> None:
...
+    def pack(self, widget: "OWBaseWidget",
+             packer: Optional[PackerType] = None) -> dict:
         """
-        if setting.packable:
-            if hasattr(instance, setting.name):
-                yield setting.name, getattr(instance, setting.name)
-            else:
-                warnings.warn("{0} is declared as setting on {1} "
-                              "but not present on instance."
-                              .format(setting.name, instance))
```
janezd, line 319: I removed this warning because I don't see a situation in which it can (reasonably) happen: how can a setting appear in a widget without being declared? Unless, of course, one adds it "manually" to the dict, or if it deletes it using delattr.

orangewidget/settings.py
```diff
@@ -719,6 +714,79 @@ def is_allowed_type(cls, tp) -> bool:
             return args[0] in (str, bool, int, float) \
                    and cls.is_allowed_type(args[1])
 
+    @classmethod
+    def check_type(cls, value, tp) -> bool:
```
janezd, line 722: How do you annotate `tp` here? It is a Union of Setting, type and ... the things you get from typing, like List or Union or Tuple ... The latter don't have a common, accessible base type, AFAIK.

janezd:
I took a deep breath, put my helmet on, and clicked Merge.

ajdapretnar:
Is it possible this just created a crash in File widget?
@thocevar found this.
Add File. Remove File. Error pops up.

PR #65, opened 2020-05-15
MessageWidget: enable messages to be clickable

### LINKED ISSUE #63: Error message not clickable
##### Orange version

3.26.0.dev

##### Expected behavior
The user can see the entire error message. If the message is too long, one can click on the message and see it.

##### Actual behavior
The error message is not shown entirely for some widgets.

##### Steps to reproduce the behavior
File - Select Columns (remove all features from attributes) - Logistic Regression (or Naive Bayes).

aturanjanin:
##### Issue
Fixes #63.

##### Description of changes
Message box was only clickable if more than one message was shown, otherwise whole message was visible only in a tooltip. Now is set to be clickable whenever there is a message.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/utils/messagewidget.py
+++ b/orangewidget/utils/messagewidget.py
@@ -376,9 +376,6 @@ class MessagesWidget(QWidget):
         #: The full (joined all messages text - rendered as html), displayed
         #: in a tooltip.
         self.__fulltext = ""
-        #: The full text displayed in a popup. Is empty if the message is
-        #: short
-        self.__popuptext = ""
         #: Leading icon
         self.__iconwidget = IconWidget(
             sizePolicy=QSizePolicy(QSizePolicy.Fixed, QSizePolicy.Fixed)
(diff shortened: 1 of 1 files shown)


orangewidget/utils/messagewidget.py
```diff
@@ -587,11 +587,8 @@ def is_short(m):
         self.__fulltext = fulltext
         self.setToolTip(self.__styled(self.__defaultStyleSheet, fulltext))
         self.anim.start(QPropertyAnimation.KeepWhenStopped)
+        self.__popuptext = fulltext if messages else ""
```
lanzagar, line 590: Wouldn't `self.__popuptext = fulltext` be enough because when the `if messages` does not apply then fulltext would be `""` anyway (line 579 above).
In that case, do we really still need `self.__popuptext` if it is the same as `self.__fulltext`?

aturanjanin: I've just updated the PR with this fixed. Thanks.

lanzagar:
Great! Simplifications are always welcome :)

PR #68, opened 2020-05-22
[ENH] Report add more possibilities to report_table

PrimozGodec:
##### Issue

`report_table` does not format decorators and graphic scenes.

##### Description of changes
- Implement that decorators are included in the report table
- Implement that graphic scenes are also rendered in the report table

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/orangewidget/report/report.py
+++ b/orangewidget/report/report.py
@@ -1,12 +1,13 @@
 import itertools
 import time
 from collections import OrderedDict, Iterable
+from typing import Optional
 
 from AnyQt.QtCore import (
-    Qt, QAbstractItemModel, QByteArray, QBuffer, QIODevice
-)
-from AnyQt.QtGui import QColor, QBrush
-from AnyQt.QtWidgets import QGraphicsScene, QTableView, QMessageBox
+    Qt, QAbstractItemModel, QByteArray, QBuffer, QIODevice,
+    QSize)
+from AnyQt.QtGui import QColor, QBrush, QIcon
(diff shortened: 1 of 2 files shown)


orangewidget/report/report.py
```diff
@@ -590,7 +598,22 @@ def get_html_img(scene):
...
+
+
+def get_icon_html(icon: QIcon, size: QSize) -> str:
+    print(size)
+    print(size.isValid())
+    byte_array = QByteArray()
+    buffer = QBuffer(byte_array)
+    buffer.open(QIODevice.WriteOnly)
+    icon.pixmap(size).save(buffer, "PNG")
+    img_encoded = byte_array.toBase64().data().decode("utf-8")
```
PrimozGodec, line 615: Tests fail since this line returns an empty string. I tried to reproduce this behavior locally but I can't even when downgrading Pyqt to a version equal to the version that Tavis uses (I tested on python3.7). @ales-erjavec do you have any idea what is the problem here?

ales-erjavec: Maybe
```diff
     buffer.open(QIODevice.WriteOnly)
     icon.pixmap(size).save(buffer, "PNG")
+    buffer.close()
     img_encoded = byte_array.toBase64().data().decode("utf-8")
```

Also ss `icon.pixmap(size)` valid?

PrimozGodec: Thank you.

PR #69, opened 2020-05-23
ContextHandler: Implement schema-only context settings

janezd:
##### Issue

Implements schema-only context settings (resolves https://github.com/biolab/orange3/issues/4753)

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/orangewidget/settings.py
+++ b/orangewidget/settings.py
@@ -36,7 +36,6 @@ import os
 import logging
 import pickle
 import pprint
-import time
 import warnings
 from operator import itemgetter
 from typing import Any, Optional, Tuple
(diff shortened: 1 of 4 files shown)


orangewidget/settings.py
```diff
@@ -795,9 +790,15 @@ def update_defaults(self, widget):
...
-        ids = {id(c) for c in globs}
-        globs += (c for c in widget.context_settings if id(c) not in ids)
-        globs.sort(key=lambda c: -c.time)
+        new_contexts = []
+        for context in widget.context_settings:
+            context = copy.deepcopy(context)
+            for setting, data, _ in self.provider.traverse_settings(data=context.values):
+                if setting.schema_only and hasattr(data, setting.name):
+                    delattr(data, setting.name)
+            if context not in globs:
```
janezd, line 799: Context used to be matched by their id's to prevent multiplying the same context. This is no longer possible because we must copy the widget's context before removing schema-only settings and merge them into global contexts.

I hope `context not in globs` will indeed work. This also required removing the `time` attribute. I shed no tears for it; I have already considered eliminating it for a long time.

orangewidget/tests/base.py
```diff
@@ -689,5 +691,25 @@ def open_widget_classes():
         yield
 
 
+@contextmanager
```
janezd, line 694: This code is moved from orangewidget/tests/test_settings_handler.py to here, so it can be shared between test files.

The only change is the additional argument `handler`.

PR #74, opened 2020-06-10
[ENH] Visual settings: saving settings, custom dialogs

VesnaT:
##### Issue

User is forced to use the predefined dialog when clicking upon the Visual Settings icon in the status bar.

##### Description of changes
- Enable using custom Visual settings dialog
- Enable using saved settings
- Remove OrderedDict

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/utils/visual_settings_dlg.py
+++ b/orangewidget/utils/visual_settings_dlg.py
@@ -3,29 +3,38 @@ from typing import List, Iterable, Tuple, Callable, Union, Dict
 from functools import singledispatch
 
 from AnyQt.QtCore import Qt, pyqtSignal as Signal
-from AnyQt.QtGui import QFont, QFontDatabase
 from AnyQt.QtWidgets import QDialog, QVBoxLayout, QComboBox, QCheckBox, \
     QDialogButtonBox, QSpinBox, QWidget, QGroupBox, QApplication, QFormLayout
 
 from orangewidget import gui
+from orangewidget.widget import OWBaseWidget
+
+KeyType = Tuple[str, str, str]
+ValueType = Union[str, int, bool]
+ValueRangeType = Union[Iterable, None]
+SettingsType = Dict[str, Tuple[ValueRangeType, ValueType]]
 
 
-class VisualSettingsDialog(QDialog):
-    """
-    A dialog for visual settings manipulation.
+class SettingsDialog(QDialog):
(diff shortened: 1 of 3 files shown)


markotoplak:
Thanks, @VesnaT, I think this PR makes it easier to understand where are the settings specified in the widgets. It also allows the dialog to be fully customizable.

@janezd, merge freely if you agree.

orangewidget/utils/visual_settings_dlg.py
```diff
@@ -110,13 +112,40 @@ def show_dlg(self):
         self.activateWindow()
 
 
+class VisualSettingsDialog(SettingsDialog):
+    """ A dialog for visual settings manipulation, that can be uses along
```
janezd, line 116: uses -> used

orangewidget/widget.py
```diff
@@ -611,16 +609,11 @@ def _():
...
-                dlg = VisualSettingsDialog(self)
-                dlg.initialize(self.initial_visual_settings)
-                dlg.setting_changed.connect(self.set_visual_settings)
-
                 icon = QIcon(gui.resource_filename("icons/visual-settings.svg"))
                 icon.addFile(gui.resource_filename("icons/visual-settings-hover.svg"),
                              mode=QIcon.Active)
                 b = SimpleButton(icon=icon, toolTip="Set visual settings")
-                b.clicked.connect(dlg.show_dlg)
+                b.clicked.connect(self.openVisualSettingsClicked)
```
janezd, line 616: I support the change. I feared that enforcing a certain form of dialog is to rigid, but I forgot to write so.

In practice, the signal `openVisualSettingsClicked` will most often connect to a method that will do exactly the same as the lines you removed. Perhaps you could put them into some kind of utility function? I guess the function that contains the above lines could be a static function of `SettingsDialog`, something like

```python
@staticmethod
def attach(widget):
    assert widget.initial_visual_settings is not None
    dlg = SettingsDialog(widget, widget.initial_visual_settings)
    dlg.settings_changed.connect(widget.set_visual_settings)
    widget.openVisualSettings.clicked.connect(dlg.show_dlg)
```

I obviously haven't tested this. And I may have misread/misunderstood everything.

VesnaT: Have you seen the __init__ in VisualSettingsDialog? It does what you proposed in the `attach`.

PR #78, opened 2020-06-17
[ENH] signals: Explicit insert/remove handlers

ales-erjavec:
##### Issue

Ref: https://github.com/biolab/orange-canvas-core/issues/86

Requires https://github.com/biolab/orange-canvas-core/pull/161 to be merged and released.

##### Description of changes

Add a new `MultiInput` description to managed explicitly ordered sequence of (multiple) inputs.

i.e. 
```python
...
class Inputs:
    foo = MultiInput(...)
...
@Inputs.foo
def set_foo(self, index, foo):
   self.foo[index] = foo

@Inputs.foo.insert
def insert_foo(self, index, foo):
    self.foo.insert(index, foo)

@Inputs.foo.remove
def remove_foo(self.index)
    self.foo.pop(index)
```

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/doc/source/orange-demo/orangedemo/OWLearningCurveA.py
+++ b/doc/source/orange-demo/orangedemo/OWLearningCurveA.py
@@ -1,4 +1,4 @@
-from collections import OrderedDict
+from typing import List, Optional, Sequence
 
 import numpy
 from AnyQt.QtWidgets import QTableWidget, QTableWidgetItem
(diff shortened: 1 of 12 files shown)


ales-erjavec:
Unfortunately this solution would not properly handle undo/redo in canvas: https://github.com/biolab/orange-canvas-core/pull/118. It can only 'communicate' link appends not inserts.

markotoplak:
The concept of the solution looks good to me.

@ales-erjavec, is this still RFC or could it be merged after canvas-core release?

PR #96, opened 2020-09-18
gui: Make spinboxes draggable, make postponing synchronization of value definitive (deprecate callbackOnReturn)

irgolic:
##### Description of changes
I've always found these fun to use.
ezgif-3-9003e6b04f83

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -606,15 +606,15 @@ class SpinBoxMixin:
         return super().eventFilter(obj, event)
 
 
-class SpinBoxWFocusOut(SpinBoxMixin, QtWidgets.QSpinBox):
+class SpinBox(SpinBoxMixin, QtWidgets.QSpinBox):
     """
     A class derived from QSpinBox, which postpones the synchronization
-    of the control's value with the master's attribute until the control looses
-    focus or user presses Tab when the value has changed.
+    of the control's value with the master's attribute until the control loses
+    focus, and adds click-and-drag to change value functionality.
     """
 
 
-class DoubleSpinBoxWFocusOut(SpinBoxMixin, QtWidgets.QDoubleSpinBox):
+class DoubleSpinBox(SpinBoxMixin, QtWidgets.QDoubleSpinBox):
     """
     Same as :obj:`SpinBoxWFocusOut`, except that it is derived from
     :obj:`~QDoubleSpinBox`"""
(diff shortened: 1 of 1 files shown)


ajdapretnar:
Can we in this case turn off scrollable spin box? I much prefer to make a conscious interaction (press, hold, drag), than scrolling down the widget just to find out I have accidentally changed some parameters (and initiated yet another recomputation).

irgolic:
> Can we in this case turn off scrollable spin box? I much prefer to make a conscious interaction (press, hold, drag), than scrolling down the widget just to find out I have accidentally changed some parameters (and initiated yet another recomputation).

@ajdapretnar Like so?

irgolic:
I've noticed that postponing synchronization of the control's value until editing is finished is broken, even on master. I'll refactor and give that a fix in this PR too.

irgolic:
@janezd, @VesnaT One thing I don't understand, how is the `callbackOnReturn` kwarg supposed to work? Shouldn't all spinboxes wait until user input is finished (e.g., the user is finished typing in a number, or the user lets go of the mouse button) before propagating? Unless you were going for something like time slice, where changes propagate as you move the sliding window (in this case, as you change the value of the spinbox). But time slice comes with controls for how often the signal is propagated.

Maybe I'm understanding this wrong, but please take a look at the changes in the `spin()` function.

janezd:
Yes, I suppose `editingFinished` would be a better signal to connect to. We can probably deprecate `callbackOnReturn` and eliminate `SpinBoxWFocusOut` and `DoubleSpinBoxWFocusOut`. I don't think there's any place where we'd like to update the UX while the user is still editing the value.

irgolic:
> Yes, I suppose `editingFinished` would be a better signal to connect to. We can probably deprecate `callbackOnReturn` and eliminate `SpinBoxWFocusOut` and `DoubleSpinBoxWFocusOut`. I don't think there's any place where we'd like to update the UX while the user is still editing the value.

Done.

One more thing though, double spinbox formats text with `float(str(text).replace(",", "."))`, it doesn't let me input commas in a double spinbox in the first place though, so this seems rather pointless. I've removed it in the latest commit, but if supporting commas as decimal delimiters is a requested feature, let me know and I'll overrride the `validate()` method (I'd rather avoid this because it'll be error prone).

markotoplak:
Please bring back SpinBoxWFocusOut and DoubleSpinBoxWFocusOut, at least temporarily, as an alias.

As it is now, non-master Orange crashes when adding widgets with this branch. Not particularly nice when you try to test stuff and who knows which strange version combinations some users will use.

PR #102, opened 2020-09-27
Rework left side for better scrollbars and static buttonsArea

irgolic:
Check this out with biolab/orange3#5013.

##### Issue
Current implementation shows/hides scrollbar each time it's shown, this is distracting.

Also, it'd be nice to have some buttons always visible in widgets (e.g., data projection controls), regardless of scroll position.

Screenshot 2020-10-02 at 23 48 00

##### Description of changes
Implements a solution similar to the one in biolab/orange-canvas-core#122, making it dynamically update the scroll area's width when scrollbars are disabled/enabled, and when the window is large enough vertically to not show the scrollbar. Therefore, the left side can always be encapsulated in a VerticalScrollArea, making `left_side_scrolling` obsolete.

Also, the buttonsArea was now docked below a scrollable controlsArea. This change deprecates `buttons_area_orientation`.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/widget.py
+++ b/orangewidget/widget.py
@@ -466,40 +466,24 @@ class OWBaseWidget(QDialog, OWComponent, Report, ProgressBarMixin,
         self.__splitter = self._Splitter(Qt.Horizontal, self)
         self.layout().addWidget(self.__splitter)
 
-    class _HiddenVerticalScrollArea(VerticalScrollArea):
-
-        def __init__(self, parent):
-            super().__init__(parent)
-            self.setFrameShape(QFrame.NoFrame)
-
-        def sizeHint(self):
-            if self.widget():
-                return self.widget().sizeHint()
-            else:
-                return super().sizeHint()
-
-        def minimumSizeHint(self):
-            if self.widget():
-                min_height = min(self.widget().minimumSizeHint().height(), 300)
-                return QSize(self.widget().minimumSizeHint().width(), min_height)
-            else:
-                return super().minimumSizeHint()
-
     def _insert_control_area(self):
+        self.left_side = gui.vBox(self.__splitter, addSpace=0)
         if self.left_side_scrolling:
-            scroll_area = self._HiddenVerticalScrollArea(self)
-            self.__splitter.addWidget(scroll_area)
-            self.left_side = gui.vBox(scroll_area, spacing=0)
-            scroll_area.setWidget(self.left_side)
+            scroll_area = VerticalScrollArea(self.left_side)
+            scroll_area.setSizePolicy(QSizePolicy.MinimumExpanding,
+                                      QSizePolicy.Preferred)
+            self.controlArea = gui.vBox(scroll_area, spacing=0)
+            scroll_area.setWidget(self.controlArea)
+            self.left_side.layout().addWidget(scroll_area)
         else:
-            self.left_side = gui.vBox(self.__splitter, spacing=0)
-        self.__splitter.setSizes([1])  # Smallest size allowed by policy
+            self.controlArea = gui.vBox(self.left_side, spacing=0)
+
         if self.buttons_area_orientation is not None:
-            self.controlArea = gui.vBox(self.left_side, addSpace=0)
             self._insert_buttons_area()
-        else:
-            self.controlArea = self.left_side
+
         if self.want_main_area:
+            self.left_side.setSizePolicy(
+                QSizePolicy.MinimumExpanding, QSizePolicy.MinimumExpanding)
             self.controlArea.setSizePolicy(
                 QSizePolicy.Fixed, QSizePolicy.MinimumExpanding)
             m = 0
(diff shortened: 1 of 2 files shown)


orangewidget/widget.py
```diff
@@ -383,34 +380,15 @@ def _insert_splitter(self):
...
-        else:
-            self.left_side = gui.vBox(self.__splitter, spacing=0)
-        self.__splitter.setSizes([1])  # Smallest size allowed by policy
+        scroll_area = VerticalScrollArea(self)
+        self.__splitter.addWidget(scroll_area)
+        self.left_side = gui.vBox(scroll_area, spacing=0,
+                                  sizePolicy=(QSizePolicy.Fixed,
+                                              QSizePolicy.MinimumExpanding))
+        scroll_area.setWidget(self.left_side)
+        scroll_area.setMinimumWidth(1)  # forces it not to stretch on initial show
```
irgolic, line 390: I assume the 
```python
self.__splitter.setSizes([1])  # Smallest size allowed by policy
``` 
line was here to make the left side not stretch when the window is resized horizontally. This worked for all widgets except Heat Map, for some reason. However, it worked correctly after collapsing and uncollapsing the left area. `scroll_area.setMinimumWidth(1)` seems to replace this just fine, but perhaps I didn't test it with enough widgets to make sure it's the definitive solution.

irgolic:
I can't figure out why Heat Map in Orange3 behaves differently from other widgets. I think everything else is fine.

I'd like to make the left side a vertical box, with a scroll area on the top and buttonsArea on the bottom. I'll need some expert help with this, though, splitters and scroll areas are weird.

ales-erjavec:
> I can't figure out why Heat Map in Orange3 behaves differently from other widgets.

Others do to (to a lesser extent). E.g. Scatter plot when extended horizontally to 2560px.
Screenshot 2020-10-02 at 13 02 03

(Heat Map has a too small sizeHint for the main area; that is why it is more pronounced there)

I think the problem is that QSplitter does not respect the QSizePolicy.Fixed size policy (QSplitter is really a bad fit for this use case here).

The left area would need to define a correct `maximumSize` (unfortunately this is not a virtual method/property) or explicitly set `splitter.setSizes([...])` to 'unbalanced' sizes e.g. `setSizes([1, QWIDGETSIZE_MAX])` in init (like it is done in `__setControlAreaVisible`)

irgolic:
@ales-erjavec That's genius, thank you so much. <3

I've gone ahead and reworked how the buttonsArea works too, adding a `want_buttons_area` option that defaults to False.

This is a bit of a vision not in the scope of this PR, but wouldn't it be nice if you didn't have to specify `want_x` options? Maybe the areas could automatically build themselves when they're first references as a parent?

At some point I'm going to go over the margins as they're inconsistent, the distance between two boxes should be equal to the controlsArea margin to the window frame.

irgolic:
Upon further inspection, I've removed `want_buttons_area` entirely. Now, if there's both a control area and main area, the control area will always have a buttons area below it.

markotoplak:
I have tested it both on Windows and MacOS. Seems to work nicely!

markotoplak:
@irgolic, I added a commit that makes scrolling possible also with widgets that lack the main area. For an example, check the Color widget. When the widget is extended, its behaviour should be the same as before, but now we can also make it smaller (then, scrollbars appear). Could you perhaps check if I unintentionally destroyed anything?

To make it work I needed to change a line in VerticalScrollArea; size hints are now computed from `sizeHint().width()` instead of `width()`. Without the fix widgets were only able to grow wide and could not be made narrower. :)

That commit does not change anything for widgets that have `resizing_enabled = False`. I thought about "overriding" that feature too, but maybe it is not a good idea. That settings also sets the window type: on a linux machine with a tiling window manager `resizing_enabled = False` widgets popup as dialogs (floating windows), while others are expanded as much as possible (tiled windows). So for very small widgets it really does make sense.

An idea, for some other PR, perhaps: if widget's sizeHint is bigger than some limit, forcefully behave as if `resizing_enabled=True`. And show a warning. What do you think, @irgolic?

irgolic:
I like the idea, but I don't like how there's a scrollbar visible by default.

Screenshot 2020-11-07 at 16 14 50

What if, given no mainArea, we set it to minimumExpanding, and limit the height of the scroll_area? That way, if the controlArea is above a certain height, it'll appear clamped, with a scrollbar, but won't let you do this with the window:

Screenshot 2020-11-07 at 16 15 42

markotoplak:
Well, good that you tried it out, I did not get any scrollbars by default on my computer. That is what makes this things fun!

I did not quite understand how you'd solve this one, but I do agree with your proposal. Makes sense not to be able to make it all too small. If you have an idea how to do it, just proceed.

Then, we could probably forget about width testing. I think it only came up as an issue once.

PR #119, opened 2020-12-18
[DOC] Unify Getting started section

### LINKED ISSUE #118: [DOC] Widget Development: join Getting Started sections
Getting Started in Widget Development documentation has two separate sections. Make it one section.
https://orange3.readthedocs.io/projects/orange-development/en/latest/

ajdapretnar:
##### Issue

Fixes #118.

##### Description of changes
Unify Getting started section.
Update toolbox image.
Small fixes.

##### Includes
- [ ] Code changes
- [ ] Tests
- [X] Documentation


irgolic:
Cool, thanks!

One thing, the `widgettoolbox.png` image doesn't render well for me. It's got a yellowish tint, and the text looks as if it's aliasing slightly. Also, the toolbox is open by default, so perhaps the screenshot should reflect that.

ajdapretnar:
Github preview doesn't show a yellow tint. Could it be your/my monitor?
It would always alias, because of how GIMP scales images. Nothing I can do (unless you can suggest a better scaling tool).
I can replace this with an open toolbox, but the limitation is the size (one would see only Data pane, not the others). Not sure which is better: compact with all widgets or expanded with just the first couple of widgets. I guess the default, but please confirm.

irgolic:
Go for default (sorry for late reply).

ajdapretnar:
Fixed.

irgolic:
It's cool :)

I think we should default to screenshots with shadows, though. Better yet, generate documentation screenshots automatically per-platform, so you're always looking at a screenshot of your platform's version of Orange.

ajdapretnar:
Voting very much against shadows in screenshots. As for automatic generation, I'm in favour, but we have to consider some caveats (text to image).

PR #126, opened 2021-01-29
Skip minimum size check if scrollable

irgolic:
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


markotoplak:
I think we should still tests minimum size. Widgets could also prevent resizing due to strangely configured mainAreas or an absurdly big buttonAreas.

PR #127, opened 2021-02-01
widget: Stretch controlArea to buttonsArea

irgolic:
##### Issue
Fixes https://github.com/biolab/orange3/issues/5225

##### Description of changes
You're right @markotoplak, this works well for all widgets :D
Although, some more testing would be nice. Combine this PR with https://github.com/biolab/orange3/pull/5013

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -2854,10 +2854,13 @@ class VerticalScrollArea(QScrollArea):
 
     def resizeEvent(self, event):
         sb = self.verticalScrollBar()
-        if sb.minimum() == sb.maximum():
+        isTransient = sb.style().styleHint(QStyle.SH_ScrollBar_Transient, widget=sb)
+
+        if isTransient or sb.minimum() == sb.maximum():
             self.setViewportMargins(0, 0, 0, 0)
         else:
             self.setViewportMargins(0, 0, 5, 0)
+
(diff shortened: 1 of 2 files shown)


markotoplak:
Originally this PR did not work for me. I tried to fix it: I think we need to set stretch only after we add the object. :)

But I am not sure this works that well with widgets with no expanding elements. Here is a scatterplot, when space is constrained:

image

And if I expand it, is starts looking a bit strange:

image

What do you think?

irgolic:
You're right, the old version doesn't work for me either. Not sure how I messed up that commit, thanks for the fix.

But you're right, it doesn't look good with the stretch upon expanding the window. What if I make this into an obscure GUI layout option and discourage it in general?

markotoplak:
Oh, the above screenshots were done on biolab/orange3#5013.

I if do them on master, I get normal behavior (see below). So this PR is OK, I think, just biolab/orange3#5013 branch will have to be fixed by adding a rubber, as there was in the master, after controls in widgets.

image

markotoplak:
We talked about this with @irgolic. He is going to solve the problem differently.

irgolic:
I think with this commit, we're good.

PR #129, opened 2021-02-03
VerticalScrollArea: Fix transient scrollbar bug

irgolic:
##### Issue
The transient scrollbar wouldn't disappear, except upon flicking AlwaysOn scrollbar style.

##### Description of changes
Avoid instantiating a new scrollbar (which is what causes the issue apparently), and use an eventFilter instead of subclass. My guess is, setting a custom scrollBar does not properly call its private `setTransient` method the same way it does ordinarily.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


orangewidget/gui.py
```diff
@@ -2833,24 +2833,18 @@ class VerticalScrollArea(QScrollArea):
     needs to scroll horizontally: it adapts its width to the contents.
     """
 
-    class ScrollBar(QScrollBar):
```
VesnaT, line 2836: I guess `QScrollBar` can now be removed from imports.

VesnaT (approved):

PR #130, opened 2021-02-05
Widget sizeHint, button/box sizePolicy fixes

irgolic:
##### Issue
Widgets aren't visually consistent. It'll take a little bit of work in Orange3 and add-ons, but once we're there, everything will be nice. Most of the changes I made in the Orange3 PR simplify the code, I hope that's the case for addons too.

##### Description of changes
See commit history.

This was written on macOS 10.15.7, working with both QMacStyle and Plastic. Hoping Windows and Linux look ok.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -231,10 +231,11 @@ def miscellanea(control, box, parent,
     :type sizePolicy: QSizePolicy
     """
     for prop, val in kwargs.items():
-        if prop == "sizePolicy":
-            control.setSizePolicy(QSizePolicy(*val))
+        method = getattr(control, "set" + prop[0].upper() + prop[1:])
+        if isinstance(val, tuple):
+            method(*val)
         else:
-            getattr(control, "set" + prop[0].upper() + prop[1:])(val)
+            method(val)
     if disabled:
         # if disabled==False, do nothing; it can be already disabled
         control.setDisabled(disabled)
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -341,7 +341,7 @@ def rubber(widget):
     widget.layout().addStretch(100)
 
 
-def widgetBox(widget, box=None, orientation=Qt.Vertical, margin=None, spacing=4,
+def widgetBox(widget, box=None, orientation=Qt.Vertical, margin=None, spacing=None,
               **misc):
     """
     Construct a box with vertical or horizontal layout, and optionally,
@@ -378,7 +378,8 @@ def widgetBox(widget, box=None, orientation=Qt.Vertical, margin=None, spacing=4,
         if margin is None:
             margin = 0
     setLayout(b, orientation)
-    b.layout().setSpacing(spacing)
+    if spacing is not None:
+        b.layout().setSpacing(spacing)
     b.layout().setContentsMargins(margin, margin, margin, margin)
     miscellanea(b, None, widget, **misc)
     return b
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -304,11 +304,13 @@ def _addSpace(widget, space):
         Otherwise, the default space is inserted by calling a :obj:`separator`.
     :type space: bool or int
     """
-    if space:
-        if type(space) == int:  # distinguish between int and bool!
-            separator(widget, space, space)
-        else:
-            separator(widget)
+    warnings.warn("'addSpace' has been deprecated. Use gui.separator instead.",
+                  DeprecationWarning, stacklevel=4)
+    # if space:
+    #     if type(space) == int:  # distinguish between int and bool!
+    #         separator(widget, space, space)
+    #     else:
+    #         separator(widget)
 
 
 def separator(widget, width=4, height=4):
@@ -368,7 +370,7 @@ def widgetBox(widget, box=None, orientation=Qt.Vertical, margin=None, spacing=4,
         if isinstance(box, str):
             b.setTitle(" " + box.strip() + " ")
         if margin is None:
-            margin = 7
+            margin = 4
     else:
         b = QtWidgets.QWidget(widget)
         b.setContentsMargins(0, 0, 0, 0)
@@ -377,7 +379,6 @@ def widgetBox(widget, box=None, orientation=Qt.Vertical, margin=None, spacing=4,
     setLayout(b, orientation)
     b.layout().setSpacing(spacing)
     b.layout().setContentsMargins(margin, margin, margin, margin)
-    misc.setdefault('addSpace', bool(box))
     miscellanea(b, None, widget, **misc)
     return b
 
@@ -1178,7 +1179,6 @@ def radioButtons(widget, master, value, btnLabels=(), tooltips=None,
         appendRadioButton(bg, lab, tooltip=tooltips and tooltips[i], id=i + 1)
     connectControl(master, value, callback, bg.group.buttonClicked[int],
                    CallFrontRadioButtons(bg), CallBackRadioButton(bg, master))
-    misc.setdefault('addSpace', bool(box))
     miscellanea(bg.group, bg, widget, **misc)
     return bg
 
@@ -1233,7 +1233,6 @@ def appendRadioButton(group, label, insertInto=None,
     if addToLayout:
         dest = insertInto or group
         dest.layout().addWidget(w, stretch)
-        _addSpace(dest, addSpace)
     return w
 
 
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -909,10 +909,6 @@ def checkBox(widget, master, value, label, box=None,
     cbox.toggled[bool].connect(cbox.makeConsistent)
     cbox.makeConsistent(value)
     miscellanea(cbox, b, widget, **misc)
-
-    # QTBUG-2699
-    cbox.setAttribute(Qt.WA_LayoutUsesWidgetRect)
-
     return cbox
 
 
@@ -1112,10 +1108,6 @@ def button(widget, master, label, callback=None, width=None, height=None,
         button.clicked.connect(callback)
 
     miscellanea(button, None, widget, **misc)
-
-    # QTBUG-2699
-    button.setAttribute(Qt.WA_LayoutUsesWidgetRect)
-
     return button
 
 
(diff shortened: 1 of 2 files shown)


ajdapretnar:
I don't think Time Slice has this option anymore.

orangewidget/widget.py
```diff
@@ -552,7 +552,7 @@ def set_basic_layout(self):
         if not self.resizing_enabled:
             self.layout().setSizeConstraint(QVBoxLayout.SetFixedSize)
 
-        self.want_main_area = self.want_main_area or self.graph_name
+        self.want_main_area = self.want_main_area
```
irgolic, line 555: Remove

orangewidget/widget.py
```diff
@@ -500,6 +500,7 @@ def _insert_control_area(self):
         if self.buttons_area_orientation is not None:
             self._insert_buttons_area()
             self.buttonsArea.layout().setContentsMargins(*m)
+            m = m[0], m[1], m[2], m[3] - 2
```
irgolic, line 503: - Fixup to prev commit
- Add comment

orangewidget/widget.py
```diff
@@ -963,6 +963,30 @@ def __updateSavedGeometry(self):
             # Note: This should always be stored as bytes and not QByteArray.
             self.savedWidgetGeometry = bytes(self.saveGeometry())
 
+    def sizeHint(self):
+        sh = QSize()
+        width = 4
```
irgolic, line 968: Add comment

orangewidget/gui.py
```diff
@@ -304,11 +304,13 @@ def _addSpace(widget, space):
         Otherwise, the default space is inserted by calling a :obj:`separator`.
     :type space: bool or int
     """
-    if space:
-        if type(space) == int:  # distinguish between int and bool!
-            separator(widget, space, space)
-        else:
-            separator(widget)
+    warnings.warn("'addSpace' has been deprecated. Use gui.separator instead.",
+                  DeprecationWarning, stacklevel=3)
```
irgolic, line 312: Better deprecate

orangewidget/widget.py
```diff
@@ -962,6 +965,31 @@ def __updateSavedGeometry(self):
...
+        if self.want_message_bar:
+            msh = self.statusBar().sizeHint()
+            height += msh.height()
+        if self.want_control_area:
+            csh = self.controlArea.sizeHint()
+            width += csh.width()
+            height += csh.height()
+            if self.buttons_area_orientation:
+                bsh = self.buttonsArea.sizeHint()
+                height += bsh.height()
```
markotoplak, line 983: This sizeHint computation is quite complex. Why wouldn't superclass sizeHint() work if you then:
- for width, subtract mainArea's sizeHint width?
- for height, do max(height, 500)?

orangewidget/widget.py
```diff
@@ -962,6 +965,31 @@ def __updateSavedGeometry(self):
             # Note: This should always be stored as bytes and not QByteArray.
             self.savedWidgetGeometry = bytes(self.saveGeometry())
 
+    def sizeHint(self):
```
markotoplak, line 968: I understand why this would be useful for graphs, but could you please explain why is this method useful if you don't have mainArea?

Also, I do not see a way to get the original Qt behavior back. Of course, I could reimplement sizeHint and return some constants, but that is something else. What you suggest might be a bad heuristic for some specific widget, therefore I'd like a way to disable it.

orangewidget/gui.py
```diff
@@ -230,16 +231,37 @@ def miscellanea(control, box, parent,
     :param sizePolicy: the size policy for the box or the control
     :type sizePolicy: QSizePolicy
     """
+    if addSpace is not None:
+        warnings.warn("'addSpace' has been deprecated. Use gui.separator instead.",
+                      DeprecationWarning, stacklevel=3)
     for prop, val in kwargs.items():
-        if prop == "sizePolicy":
-            control.setSizePolicy(QSizePolicy(*val))
+        if prop.startswith('box_'):
```
markotoplak, line 238: For me, these are too many shortcuts. How often do you use these in the widget?

orangewidget/gui.py
```diff
@@ -230,16 +231,37 @@ def miscellanea(control, box, parent,
...
+            prop = prop[4:]
+            target = box
+        elif prop.startswith('parent_'):
+            prop = prop[7:]
+            target = parent
         else:
-            getattr(control, "set" + prop[0].upper() + prop[1:])(val)
+            target = control
+
+        dashsplit = prop.split('_')
```
markotoplak, line 247: No. If you are not using this at least 100x in Orange widgets, no. If we are, we need a better solution.

markotoplak:
While I do agree that this branch combined with the PR in Orange3 makes Orange look nicer, I can not merge this. For me, there is too much new additional magic in miscellanea (allowing `box_` and `parent_` settings) and complicated and non-avoidable magical widget `sizeHint`.

Mac-specific changes also bother me somewhat, but if that is really needed to fix strange alignments, well, then fine. These I could actually merge...

@ales-erjavec, do you have any time to review this? If you say these code patterns are fine, then I'll accept them.

markotoplak:
I can merge this PR if:
1. There is an option to disable sizeHint.
2. The new `box_`, `parent_`, multiple `_` magic gets removed from miscelanea. Yes, sometimes we will have to use multiple lines of code instead of additional parameters... But if we could survive without these before, they do not seem to be needed often.

orangewidget/widget.py
markotoplak: Since I am asking you to make this sizeHint an option... Perhaps you could also expose the 1.1 ratio? Perhaps have an attribute defining the ratio that defaults to 1.1 and then allowing the widget to set it differently? And then, if it is False (or None, I do not know) it would just return `super().sizeHint()`?

Some context: current defaults (and fixed implementation) could be problematic for widgets with two vertically arranged graphs in the main area: I have some, Hyperspectra and Preprocess Spectra.

orangewidget/gui.py
```diff
@@ -1113,6 +1124,12 @@ def button(widget, master, label, callback=None, width=None, height=None,
     :rtype: QPushButton
     """
     button = buttonType(widget)
+    if button.style().metaObject().className() == 'QMacStyle':
+        btnpaddingbox = vBox(widget, margin=0, spacing=0)
+        separator(btnpaddingbox, 0, 4)  # lines up with a WA_LayoutUsesWidgetRect checkbox
+        button.outer_box = btnpaddingbox
```
ales-erjavec, line 1130: Why would you assume anything related to checkbox here?

ales-erjavec:
> @ales-erjavec, do you have any time to review this? If you say these code patterns are fine, then I'll accept them.

I don't like it.

I already cannot reason about what the existing code does. This would make it even worse.
But then again I advise against using `gui` at all and if this fixes existing uses (for Orange only?) then go ahead.

If you do remove `addSpace` (good) why then add `addSpaceBefore` (bad)?

I would keep some changes that fix layouts locally, I.e. the one that removes *near global* `WA_LayoutUsesWidgetRect` and moves it to the function where it is required, margins/spacings in created widgets/layouts, ...

orangewidget/gui.py
```diff
@@ -1726,19 +1764,29 @@ def do_commit():
...
                                callback=checkbox_toggled, tooltip=auto_label)
     if _is_horizontal(orientation):
-        b.layout().addSpacing(10)
+        w = b.style().pixelMetric(QStyle.PM_CheckBoxLabelSpacing)
+        separator(b, w, 0)
     cb.setSizePolicy(QSizePolicy.Preferred, QSizePolicy.Preferred)
 
     b.button = btn = VariableTextPushButton(
         b, text=label, textChoiceList=[label, auto_label], clicked=do_commit)
+    btn.setMaximumHeight(btn.sizeHint().height())
```
ales-erjavec, line 1782: Can you try setting `QSizePolicy.Maximum` vertical size policy here instead? the button's size hint can change at runtime (app font/style change, ...).

orangewidget/gui.py
```diff
@@ -230,16 +231,37 @@ def miscellanea(control, box, parent,
...
+            method(*val)
+        else:
+            method(val)
     if disabled:
         # if disabled==False, do nothing; it can be already disabled
         control.setDisabled(disabled)
     if tooltip is not None:
         control.setToolTip(tooltip)
+    if box and parent and box is not parent and parent.layout() and parent.layout().spacing() == -1:
+        parent.layout().setSpacing(8)
```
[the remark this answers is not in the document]
irgolic: MacStyle is weird. It doesn't use constant spacing, it uses this weird matrix to calculate spacing between two controls, depending on which one precedes the other. https://doc.qt.io/archives/qq/qq23-layouts.html

This works great if you only use control elements. But as soon as you nest controls in boxes (e.g., label + combo), the margins and spacing get really big. Our solution for this is to set the spacing on the box (if no spacing has yet been set), if you add a nested control. For this to work right, `Qt.WA_LayoutUsesWidgetRect` should be set on all the controls in the parent box. 

... That was our rationale, but I've removed it, and I think we'll live without it.

orangewidget/gui.py
```diff
@@ -367,17 +402,21 @@ def widgetBox(widget, box=None, orientation=Qt.Vertical, margin=None, spacing=4,
         b = QtWidgets.QGroupBox(widget)
         if isinstance(box, str):
             b.setTitle(" " + box.strip() + " ")
+            if widget and widget.layout() and \
+                    isinstance(widget.layout(), QtWidgets.QVBoxLayout) and \
+                    not widget.layout().isEmpty():
+                misc.setdefault('addSpaceBefore', True)
```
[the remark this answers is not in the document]
irgolic: `addSpaceBefore` replaced `addSpace`, it's there to put separators between boxes, but not before the first box. I could make it private to the gui package, then you could disable it with `addToLayout=False`.

orangewidget/gui.py
irgolic: Autocommit is a button next to a checkbox. Generally you want all buttons in a row (QHBoxLayout) at the same height, so it's set globally.

Without this, still keeping `Qt.WA_LayoutUsesWidgetRect` will misalign the checkbox and button.

orangewidget/gui.py
irgolic: I've removed the unnecessary magic.

markotoplak: Thanks, now even I can understand it. :)

PR #136, opened 2021-03-05
Add automated summaries

### LINKED ISSUE #109: set_output_summary: Use summary when no description
@PrimozGodec wrote in https://github.com/biolab/orange3/pull/4904

> Note: self.info.set_output_summary(summary, summary) is used since in the cases with string summary and empty description, the description is not set automatically to same value than it is done for int summary. I suggest merging this PR as it is and unifying those behaviours in the widget base. I would be happy to submit the PR.

janezd:
#### Issue

Provide automated summaries for inputs and outputs. Ref: https://github.com/biolab/orange3/issues/5108.

For proper effect, must be merged with https://github.com/biolab/orange3/pull/5308, but nothing should break if not.

Also: resolves #109.

#### Description of changes

- Outputs are intercepted on calls to `send` (only new-style signals are supported; this is intentionally as the "old-style" are deprecated for many many years)
- Inputs are intercepted by using the existing decorator that marks inputs

Auto-summary is generated if
- the signal's `auto_summary` flag is not explicitly set to `False`,
- the signal type has a registered summarizing function.

Summarizing functions are registered by adding another type to single-dispatched `orangewidget.utils.signals.summarize`, which must return an instance of `orangewidget.utils.signals.PartialSummary`, e.g.

```python
@summarize.register(Table)
def summarize(data):
    return PartialSummary(data.approx_len(), format_summary_details(data))
```

The summary (first argument), can be `None`, `int` or rich text `str` (it has to be html-safe!), and the second is `None` (if also the first is `None`) or a rich text `str`.

**(Non-consequential?) imcompatibility**: if a widget already implement sets the output summary, the one who acts last will override the one who's first. That is, either nothing will change or the summary will (probably) improve.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/utils/signals.py
+++ b/orangewidget/utils/signals.py
@@ -1,9 +1,12 @@
 import copy
+import html
 import itertools
 import warnings
 from functools import singledispatch
 import inspect
 from typing import NamedTuple, Union, Optional
+
+from AnyQt.QtCore import Qt
 
 from orangecanvas.registry.description import (
     InputSignal, OutputSignal, Single, Multiple, Default, NonDefault,
@@ -27,6 +30,24 @@ def base_summarize(_) -> PartialSummary:
 
 
 summarize = singledispatch(base_summarize)
+
+SUMMARY_STYLE = """
+<style>
+    ul {
+        margin-left: 10px;
+        margin-top: 2px;
+        -qt-list-indent:0
+    }
+
+    li {
+        margin-top: 3px;
+    }
+
+    th {
+        text-align: right;
+    }
+</style>
+"""
 
 
 def can_summarize(type_, name):
@@ -339,21 +360,28 @@ class WidgetSignalsMixin:
             if isinstance(summary, int):
                 return StateInfo.format_number(summary)
             if isinstance(summary, str):
-                return summary
+                return html.escape(summary).replace("\n", "<br/>")
             raise ValueError("summary must be None, string or int; "
                              f"got {type(summary).__name__}")
 
         def format_detail(partial):
             if partial.summary is None:
                 return "-"
-            else:
-                return str(partial.details or partial.summary)
+            details = str(partial.details or partial.summary)
+            return html.escape(details).replace("\n", "<br/>")
 
         def join_multiples(partials):
             if not partials:
                 return "-", "-"
-            return ",".join(map(format_short, partials.values())), \
-                   "\n+\n".join(map(format_detail, partials.values()))
+            shorts = ",".join(map(format_short, partials.values()))
+            if len(partials) == 1:
+                details = format_detail(next(iter(partials.values())))
+            else:
+                details = "<ul>" \
+                          + "".join(f"<li>{format_detail(partial)}</li>"
+                                    for partial in partials.values()) \
+                          + "</ul>"
+            return shorts, details
 
         if partial_summary.summary is None:
             if id in summaries[name]:
@@ -363,16 +391,17 @@ class WidgetSignalsMixin:
 
         if not any(summaries.values()):
             summary, detail = empty_obj, ""
-
-        elif len(summaries) == 1:
-            # If there is just one output, skip the empty line and signal name
-            summary, detail = join_multiples(next(iter(summaries.values())))
         else:
             summary, details = zip(*map(join_multiples, summaries.values()))
             summary = " | ".join(summary)
-            detail = "\n".join(f"\n{name}:\n{detail}"
-                               for name, detail in zip(summaries, details))
-        setter(summary, detail)
+            detail = "<hr/><table>" \
+                     + "".join(f"<tr><th>{name}: </th><td>{detail}</td></tr>"
+                               for name, detail in zip(summaries, details)) \
+                     + "</table>"
+        if detail:
+            setter(summary, SUMMARY_STYLE + detail, format=Qt.RichText)
+        else:
+            setter(summary)
 
 
 class AttributeList(list):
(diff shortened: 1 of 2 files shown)


janezd:
@irgolic, if you confirm the design (you'll need this PR and https://github.com/biolab/orange3/pull/5308/files, but you can easily try other widgets, too), I'll remove summaries from other widgets, too.

irgolic:
```error
-------------------------- AttributeError Exception ---------------------------
Traceback (most recent call last):
  File "/Users/rafael/orange/orange3/Orange/widgets/data/owfile.py", line 339, in load_data
    error = self._try_load()
  File "/Users/rafael/orange/orange3/Orange/widgets/data/owfile.py", line 384, in _try_load
    self.apply_domain_edit()  # sends data
  File "/Users/rafael/orange/orange3/Orange/widgets/data/owfile.py", line 518, in apply_domain_edit
    self.Outputs.data.send(table)
  File "/Users/rafael/orange/orange-widget-base/orangewidget/utils/signals.py", line 204, in send
    self.widget.set_partial_output_summary(summarize(value), self.name, id)
  File "/Users/rafael/orange/orange-widget-base/orangewidget/utils/signals.py", line 327, in set_partial_output_summary
    self._update_summary(
  File "/Users/rafael/orange/orange-widget-base/orangewidget/utils/signals.py", line 358, in _update_summary
    if partial_summary.summary is None:
AttributeError: 'tuple' object has no attribute 'summary'
-------------------------------------------------------------------------------
```
This is what I get.

janezd:
Oops, I forgot to push the PR in orange3 repo. Please pull https://github.com/biolab/orange3/pull/5308/commits/087a2aeb82000754bc1d05c2b4609771146a5e52 again.

janezd:
Please check the tooltip for multiple inputs in Concatenate. It is ... acceptable, but any suggestions are welcome.

janezd:
I actually solved it with a bit of HTML.

Screenshot 2021-03-06 at 10 38 24
Screenshot 2021-03-06 at 10 38 39
Screenshot 2021-03-06 at 11 21 29

irgolic:
That looks amazing actually.

Screenshot 2021-03-06 at 18 39 22

Now that you've changed it to a form-type layout, the indents on a MultiInput (dots) may be redundant.

Also, the top line in the message is redundant. That might be tricky to change though, as warnings use the same MessageWidget, and there, the top line is not redundant.

janezd:
I've made a few more changes to the tooltip. As you've seen, it's html with css. Please edit at will, you're much better at this than I am.

Skipping space after comma was intentional; though grammatically incorrect, I wanted to keep those numbers closer. But feel free to add space.

I like this so much that I have already changed all widgets (= removed the summary-related code). The only widget I left is `Table`, which currently shows the input related to the currently shown table. We can decide to keep it, or remove this as well. All other widget look well (in my opinion) with automated summaries.

janezd:
> Also, the top line in the message is redundant. That might be tricky to change though,

No, I add it myself, specifically in the tool tip. I think it looks nicer, but you can remove it if you think it's better without.

irgolic:
What do you think of delimiting with just a space?
Screenshot 2021-03-09 at 17 38 22

I've come up with two variants:

#### V1
Screenshot 2021-03-09 at 19 19 29
Screenshot 2021-03-09 at 19 19 20

#### V2
Screenshot 2021-03-09 at 19 43 35
Screenshot 2021-03-09 at 19 43 29

Thoughts?

> > Also, the top line in the message is redundant. That might be tricky to change though,

> No, I add it myself, specifically in the tool tip. I think it looks nicer, but you can remove it if you think it's better without.

I would prefer to remove this (the brief message never elides, and the in/out icon doesn't render well in text). But, I can't find where you add this, could you point me to it (if it's simple to remove)?

Also, something else I'd love to do but can't figure out how to do properly is add locale-dependent thousands separators (10242 -> 10,242). It should be as simple as `f{len(data):n}`, but my locale defaults to `C/UTF-8/C/C/C/C` instead of something like `en_US.UTF-8`. What's our stance on locale-dependent formatting, should I just use commas for separators?

irgolic:
Or maybe even this:

Screenshot 2021-03-09 at 20 39 22

Now imagine the entries being selectable. On click, a transient Data Table window opens.

@janezd, let me know which one of these I should push into the PR.

janezd:
Oh, I'm sorry about the line. I thought that you disliked the <hr>, not what's above it. :))) No, this is added elsewhere. And, yes, it's redundant, the icon doesn't look nice, and the line would better be removed.

I really like V1, but bear in mind that the table's name can be `"untitled"`, in which case it's currently not shown. Perhaps show it as **(untitled)**? Bullets in V2 are redundant, and I don't line centering in combination with right-aligned left column.

As for locale separators, I've no strong opinion, but leaning towards commas because Orange is in English.

Spaces instead of commas in the summary: if you like it better, do it. It's not much of an issue, it's not very common.

irgolic:
This is what I've settled on:

Screenshot 2021-03-09 at 21 51 02

Also, can I suggest trying 'Squash and merge' for this PR? The team behind pandas default to that in their workflow.

janezd:
I left separate commits in case we'd like to revert some. But at the end, yes, squash without mercy. (My last commit's message included "to be squashed".)

> Does this version include the thousands separator for the number of instances on your machine?

Which version? With this PR, as currently on github (just my commits, without any of your changes), it doesn't show any separators.

irgolic:
> > Does this version include the thousands separator for the number of instances on your machine?
> 
> Which version? With this PR, as currently on github (just my commits, without any of your changes), it doesn't show any separators.

Ah sorry, I've pushed now.

irgolic:
I can get the separator to work if I set the locale within python code, e.g.
```python
import locale
locale.setlocale(locale.LC_ALL, 
                 locale.getdefaultlocale())
```
I suspect locale is only broken in our dev environments, where it's set to C.UTF-8. Even so, I think I'd prefer to set it to en_US.UTF-8 for everyone, for now.

PR #137, opened 2021-03-05
Better deprecate addSpace

irgolic:
##### Issue
`addSpace` deprecation warnings are thrown from within `widget.py`

##### Description of changes
- better deprecates addSpace (also catching =None)
- fixes miscellania docstring
- resolves deprecation

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -186,7 +186,7 @@ class OWComponent:
 
 
 def miscellanea(control, box, parent,
-                addToLayout=True, stretch=0, sizePolicy=None, addSpace=None,
+                addToLayout=True, stretch=0, sizePolicy=None,
                 disabled=False, tooltip=None, disabledBy=None,
                 addSpaceBefore=False, **kwargs):
     """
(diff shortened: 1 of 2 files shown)


orangewidget/gui.py
```diff
@@ -186,7 +186,7 @@ def __setattr__(self, name, value):
 
 
 def miscellanea(control, box, parent,
-                addToLayout=True, stretch=0, sizePolicy=None, addSpace=None,
+                addToLayout=True, stretch=0, sizePolicy=None,
```
janezd, line 189: Let us pray that nobody called this function with non-keyword arguments (which is indeed very unlikely as this is a very internal function, although not named as such).

Maybe add a `*, ` before `addToLayout` to ensure that the function is called wiith keyword arguments and that any code that may currently call it positional arguments will fail here and not silently misbehave latter.

irgolic: 👍

PR #138, opened 2021-03-09
gui: Make draggable spinbox text selectable, fix disabled behavior

### LINKED ISSUE #134: Changing values of disabled spins
`orangewidgets.gui.SpinBoxMixin` does not check whether the spin is enabled and thus allows the user to change values of disabled spins by clicking and dragging.

irgolic:
##### Issue
Fixes #134 

##### Description of changes
It still listens for drag events on the text, but I've made the value increase exponentially (not linearly), so it should be easy enough to select the text. Is this okay? If not, I'll remove the listener from the textbox. @borondics @janezd 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -594,7 +594,8 @@ class SpinBoxMixin:
             self.cfunc()
 
     def eventFilter(self, obj, event):
-        if not (isinstance(obj, SpinBoxMixin) or isinstance(obj, QLineEdit)):
+        if not self.isEnabled() or \
+                not (isinstance(obj, SpinBoxMixin) or isinstance(obj, QLineEdit)):
             return super().eventFilter(obj, event)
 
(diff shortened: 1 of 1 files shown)


janezd:
Heh, what we have now is that if you change the value by dragging, the text is selected, though this is not the user's intention. It doesn't bother me much, I can accept this solution. @markotoplak, @borondics?

janezd:
This doesn't fix #134, though!

Connect File -> Create Table. The second spin in Create Table wil be disabled because columns are determined by the data on the input. But the user can change the value of the disabled spin by dragging, which crashes the widget.

irgolic:
> This doesn't fix #134, though!
> 
> Connect File -> Create Table. The second spin in Create Table wil be disabled because columns are determined by the data on the input. But the user can change the value of the disabled spin by dragging, which crashes the widget.

Whoops, got caught up in the issue's tangent. Current PR should fix everything.

PR #151, opened 2021-04-21
Implement signals for Setting changes

irgolic:
##### Issue
Canvas-core is not notified of Setting changes.

##### Description of changes
Each time a Setting is changed (setattr is called), `attr_changed[str]` defined in OWComponent emits, and WidgetsScheme's `sync_node_properties` is called. The latter method emits `node_properties_changed`, defined in the corresponding canvas-core PR.

This paves the way for great UX improvements, as detailed in the canvas-core PR. Node properties should definitely be synced with canvas-core whenever they are written. That said, `settingsAboutToBePacked` should be used for lazily writing Settings.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
I haven't checked it out, but I read the commits and it seems OK. @ales-erjavec, could you review and merge, if you agree with changes?

PR #153, opened 2021-05-04
[FIX] DateTimeEditWCalendarTime: Fix QDateTime module

VesnaT:
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -2965,7 +2965,6 @@ class DateTimeEditWCalendarTime(QDateTimeEdit):
             date_time = QtCore.QDateTime.currentDateTime()
         if isinstance(date_time, QtCore.QTime):
             self.setDateTime(
-                QtCore.QDateTime(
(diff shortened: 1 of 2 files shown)


janezd:
As a punishment for my mistake, I wrote a test, and naturally found another bug: when the method is passed `QTime`, it ignores it and resets the widget to the time it currently contains (= does nothing). I fixed it (https://github.com/biolab/orange-widget-base/pull/153/commits/291d1fc9bd6fbe06fe37db9e9663d9eedd46dfe2). Please review and merge if you agree.

VesnaT:
After some further investigation I noticed: if one changes only the month or the year it does not reflect on the value, wheres changing the date or the time does. 
Besides, when changing the date, the popup closes, meanwhile changing anything the popup remains open.
I can merge the PR, because it fixes a bug, but I disagree with the inconsistency.

PR #157, opened 2021-06-19
[RFC] Introduce @delayed decorator for commit/apply

janezd:
##### Issue

An attempt to solve: https://github.com/biolab/orange3/issues/5453. See details there.

##### Description of changes

Introduce a decorator @delayed that causes the method to be assigned an attribute `delayed`, which is a thunked call of `conditional_commit`, created by `auto_commit`.

Minor change: rename `unconditional_commit`, which used to be a local function, to `conditional_commit`.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
Tests fail due to insufficient mocking. I'll fix them when/if this design is approved.

orangewidget/gui.py
```diff
@@ -1669,6 +1670,26 @@ def comboBox(widget, master, value, box=None, label=None, labelWidth=None,
     return combo
 
 
+def deferred(func):
+    @wraps(func)
+    def _commit_replacement():
+        raise RuntimeError(
+            "This function is deferred; explicitly call "
+            f"{func.__name__}.deferred or {func.__name__}.now")
```
VesnaT, line 1678: If I understand this correctly `self.commit.deferred` is equivalent to the old`self.commit` and `self.commit.now` to `self.unconditional_commit`. In that case, I would remove the `or {func.__name__}.now` from the exception message and create a separate RuntimeError for the unconditional_commit.

VesnaT (approved):
Other than the minor comment, I like the change.

orangewidget/gui.py
```diff
@@ -1786,7 +1808,19 @@ def do_commit():
...
+    if decorated:
+        commit.__now = do_commit
+        commit.__deferred = conditional_commit
+    else:
+        commit_name = next(LAMBDA_NAME) if isinstance(commit, LambdaType) else commit.__name__
+        warnings.warn(
+            f"decorate {commit_name} with @gui.deferred and "
+            f"then explicitly call {commit_name}.now or {commit_name}.deferred")
+        setattr(master, 'unconditional_' + commit_name, commit)
+        setattr(master, commit_name, conditional_commit)
```
lanzagar, line 1822: This just triggers warnings when using auto_commit without the decorator. After you add the decorator, it will not complain anymore, but could crash (possibly infrequently and in specific situations not covered by tests) if someone forgets to convert an unconditional_commit call to a .now call.
If this would still (for some transition period) create the unconditional method (with a warning to use .now instead), it would be more bulletproof wrt backwards compatibility albeit uglier.

I am fine with assuming that the decorating will be done carefully and we can leave this as is. I expect searching inside the file for unconditional_commit should be enough and we don't have (m)any cases where that method was being called from another module?

orangewidget/gui.py
janezd: Your point is, to my understanding, that previously such call resulted in the deferred call. Skipping `or {func.__name__}.now` would be the correct instruction for converting the old code, but misleading for new code(rs), where both calls are potentially correct.

I guess I'd keep the message as it is.

orangewidget/gui.py
janezd: It would also raise an exception if somebody forgets to convert deferred calls - see the comment by @VesnaT.

Unconditional commits would not be much of a problem, because they are called after handling inputs and thus tested. Conditional commits are less tested because they often happen as result of use interaction with the interface.

I thought about transitional period (by replacing the original method with deferred one instead of the one that raises exceptions), but I realised it may not make much sense: this transitional period would allow us some time for fixing mistakes that we make while decorating? It's better that incorrect new code fails when written, not half a year later. :)

As a general pattern, this would be as if you wrote new function that will for some time allow somewhat incorrect arguments (but become more strict later) to allow sloppy calls (of this new function!) to be fixed.

Let us rather assume we'll decorate properly. It shouldn't be so hard, it's just about searching for calls of this method.

janezd:
(continued from https://github.com/biolab/orange3/pull/5495#issuecomment-883265654)

@ales-erjavec, brilliant. Two widgets already suffice for counterexample. I added a test that fails and to my understanding captures the issue with my code (`w.real_apply` is called with `w2` instead of with `w` because `w2` "rebinds" the method to itself).

```
        w2 = Widget()
        w.apply.now()
        w.real_apply.assert_called_with(w)
```

You're right, there can be no source of `self` within a call of `_commit_replacement.now`, because neither `_commit_replacement` nor `now` have seen (or been bound to) `self`. The fixed code uses properties; `getter` has access to `self` and can create a unique `_commit_replacement` function for each instance, so this function can be assigned instance-specific attributes in `auto_commit`.

Am I right that this fixes the problem or is there more to it than what I understood?

I've put the changes into a new commit, but everything can be squashed when/if merging.

janezd:
@markotoplak, I added comments and made other changes that we talked about. I think this is ready for review.

... except for the consistently failing tests on Ubuntu with Python 3.6. I've no idea why, and they fail without showing any output?!

orangewidget/gui.py
```diff
@@ -1677,6 +1678,224 @@ def comboBox(widget, master, value, box=None, label=None, labelWidth=None,
...
+        # if True, data was changed while auto was disabled,
+        # so enabling auto commit must call `func`
+        dirty: bool = False
+        # A flag (counter) telling that we're within commit and
+        # super().commit() should not raise an exception
+        commit_depth: int = 0
+else:
+    class DeferredData:
+        def __init__(self):
+            self.naw = self.deferred = None
```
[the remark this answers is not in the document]
JakaKokosar: Typo?

orangewidget/gui.py
```diff
@@ -1791,7 +1791,7 @@ class DeferredData:
 else:
     class DeferredData:
         def __init__(self):
-            self.naw = self.deferred = None
```
janezd: Thonks!

PR #158, opened 2021-06-26
gui: Spinbox click and hold fix

irgolic:
##### Issue
Click-and-hold on the up/down buttons was not working properly with spinboxes.

##### Description of changes
Fix click-and-hold issue. Also change the rate at which click+drag changes the value.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -550,6 +550,7 @@ class SpinBoxMixin:
         self.preEditvalue = 0
 
         self.lineEdit().installEventFilter(self)
+        self.installEventFilter(self)
         self.editingFinished.connect(self.__onEditingFinished)
         self.valueChanged.connect(self.__onValueChanged)
 
@@ -604,7 +605,7 @@ class SpinBoxMixin:
             self.mouseStartPos = event.globalPos()
             self.preDragValue = self.value()
             self.mouseHeld = True
-        elif event.type() == QEvent.MouseMove and self.mouseHeld:
+        elif event.type() == QEvent.MouseMove and self.mouseHeld and isinstance(obj, QLineEdit):
             # do click+drag
             # override default cursor on drag
             if QApplication.overrideCursor() != cursor:
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -550,7 +550,6 @@ class SpinBoxMixin:
         self.textEditing = False
         self.preEditvalue = 0
 
-        self.installEventFilter(self)
         self.lineEdit().installEventFilter(self)
         self.editingFinished.connect(self.__onEditingFinished)
         self.valueChanged.connect(self.__onValueChanged)
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -621,11 +621,17 @@ class SpinBoxMixin:
             posVal = pos.y() if self.verticalDirection else -pos.x()
             posValStart = self.mouseStartPos.y() if self.verticalDirection else -self.mouseStartPos.x()
             diff = posValStart - posVal
+
             # these magic params are pretty arbitrary, ensure that it's still
             # possible to easily highlight the text if moving mouse slightly
             # up/down, with the default stepsize
-            valueOffset = int((diff / 25) ** 3) * self.stepSize
+            normalizedDiff = abs(diff) / 30
+            exponent = 1 + min(normalizedDiff / 10, 3)
+            valueOffset = int(normalizedDiff ** exponent) * self.stepSize
+            valueOffset = math.copysign(valueOffset, diff)
+
             self.setValue(self.preDragValue + valueOffset)
+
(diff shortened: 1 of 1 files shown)


markotoplak:
@irgolic, thanks. Could you please explain what you meant by "was not working properly"? What happened vs. what was expected?

markotoplak:
I once also heard two problem reports with some spinboxes: (1) from @borondics, that clicking on an arrow-button on a k-means number of groups spinbox increases its state by 2 instead of 1 and (2) if I am not mistaken, from @ajdapretnar, who noticed somewhere that clicking a button resulted in a correct change, which was then immediately reset. Could these be connected to your PR?

irgolic:
Sorry, I should've been more explicit. This PR aims to fix the problem reported by @ajdapretnar, wherein pressing the arrows will change the value, but releasing the mouse button will reset them to their initial value. This happens on macOS Big Sur and (to my knowledge) nowhere else. By the nature of this fix, what @borondics reported _should_ also be fixed, could you give it a spin, @borondics?

markotoplak:
Are you sure that the code now works correctly and does not contain hidden bugs similar to the one you are fixing here? Why don't you disable drags on the buttons and allow them on the text-field-like part? Is it even expected, from a user perspective, that a button allows a drag that changes values?

irgolic:
> Are you sure that the code now works correctly and does not contain hidden bugs similar to the one you are fixing here?

Of course I'm not sure :) But it's written much more safely now.

> Why don't you disable drags on the buttons and allow them on the text-field-like part?

Done

markotoplak:
> > Are you sure that the code now works correctly and does not contain hidden bugs similar to the one you are fixing here?
> 
> Of course I'm not sure :) But it's written much more safely now.
> 
> > Why don't you disable drags on the buttons and allow them on the text-field-like part?
> 
> Done

Wow, that was fast! I did not realize it is so easy, if you know where to look, at least. :D Thanks.

markotoplak:
I like the changes, especially since it makes it safer!

Now that click+drag only works on the line edit, I assume that also the code that disables/enables click and hold behavior could be removed.

orangewidget/gui.py
```diff
@@ -619,16 +618,23 @@ def eventFilter(self, obj, event):
...
             self.setValue(self.preDragValue + valueOffset)
+
         elif event.type() == QEvent.MouseButtonRelease:
             # end click+drag
             # restore default cursor on release
-            if QApplication.overrideCursor() == cursor:
+            while QApplication.overrideCursor() is not None:
                 QApplication.restoreOverrideCursor()
 
             # restore click and hold behavior
```
markotoplak, line 640: Now, that buttons are not connected to this event, could the `restore click and hold behavior` code section be removed?

markotoplak: Also, the same for `# disable click and hold behavior` a bit higher.

markotoplak:
@ajdapretnar, @borondics , does this fix the problems you were having?

ajdapretnar:
Yes, this fixes it. It works much better than before.
There's still one thing, but possibly out of scope of this PR: when I start changing the parameters in a spin box, but come back to the initial setting (value of spin box didn't change), commit is called one again. Probably not ideal. Should I make it a separate issue?

irgolic:
@markotoplak Done! Turns out the fix wasn't *that* simple, check the last commit

@ajdapretnar Should be fixed, could you check? Also fixed a bug where holding the up/down buttons would send the value every time it changed

ajdapretnar:
Works well now. Thanks!
Fixes also https://github.com/biolab/orange3-text/issues/682, because now one _has to_ focus out for the value to be sent. So effectively, commit is called only once.

markotoplak:
@ajdapretnar, thanks for finding the text input behavior change! Reading out the whole code it was obvious why it happened. But should this really fix the bug in text? I only understand that this gives the pre-PR behaviour back. Or am I mistaken?

@irgolic, for consistency, I'd filter everything scrolling connected with `isinstance(obj, QLineEdit)`. Now you do it only for moves, and I'd imagine button click + move mouse left to the text field + then up and down could have some unintended consequences. Did not run it though. :) Otherwise, looks good!

ajdapretnar:
It does not "fix" the bug as was described, but it does remove two calls.
Before: One can set the value by typing it. Press Enter to run the computation. The focus is still there, so if one focuses out, the computation is re-run.
Now: Set the value by typing. Enter focuses out and runs the computation.
For me it works well now.

markotoplak:
I was wrong before. The behaviour I was afraid of does not happen, because, as @irgolic pointed out, Qt docs say: "If mouse tracking is switched off, mouse move events only occur if a mouse button is pressed while the mouse is being moved. If mouse tracking is switched on, mouse move events occur even if no mouse button is pressed."

PR #159, opened 2021-06-29
AbstractTableModel: Refactor sort

janezd:
##### Description of changes

- Allow lexical sorting
- Expose a method for setting sort indices

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/utils/itemmodels.py
+++ b/orangewidget/utils/itemmodels.py
@@ -92,7 +92,7 @@ class AbstractSortTableModel(QAbstractTableModel):
             data = numpy.array([self.index(row, column).data()
                                 for row in range(self.rowCount())])
 
-        assert data.ndim == 1, 'Data should be 1-dimensional'
+        assert data.ndim in (1, 2), 'Data should be 1- or 2-dimensional'
         data = data[self.mapToSourceRows(Ellipsis)]
         return data
 
@@ -161,7 +161,10 @@ class AbstractSortTableModel(QAbstractTableModel):
         Return indices of sorted data. May be reimplemented to handle
         sorting in a certain way, e.g. to sort NaN values last.
         """
-        indices = numpy.argsort(data, kind="mergesort")
+        if data.ndim == 1:
+            indices = numpy.argsort(data, kind="mergesort")
+        else:
+            indices = numpy.lexsort(data.T[::-1])
         if order == Qt.DescendingOrder:
             indices = indices[::-1]
         return indices
--- a/orangewidget/utils/itemmodels.py
+++ b/orangewidget/utils/itemmodels.py
@@ -10,7 +10,7 @@ from warnings import warn
 
 from AnyQt.QtCore import (
     Qt, QObject, QAbstractListModel, QAbstractTableModel, QModelIndex,
-    QItemSelectionModel, QMimeData, QT_VERSION
+    QItemSelectionModel, QMimeData
 )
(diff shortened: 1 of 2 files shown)


orangewidget/utils/itemmodels.py
```diff
@@ -218,10 +211,25 @@ def sort(self, column: int, order: Qt.SortOrder = Qt.AscendingOrder):
...
+                data = numpy.arange(self.rowCount())
+            elif data.dtype == object:
+                data = data.astype(str)
+
+            # _sortColumnData returns data in its currently shown order
+            # _argSortData thus returns an array a, in which a[i] is the row
+            # number (in the current view) to put to line i
+            # mapToSourceRows maps these indices back to original data
+            # @kernc, tbanks for this joke. :) Couldn't _sortColumnData simply
+            # return the original data, saving us from double mapping?
```
ales-erjavec, line 230: No. This is done to ensure stable sort order on consecutive calls.

janezd: Thanks! I'll document that. Seeing the code, I had to convince myself it works at all, then I saw the trick but didn't understand why.

orangewidget/utils/itemmodels.py
```diff
@@ -180,33 +183,21 @@ def sort(self, column: int, order: Qt.SortOrder = Qt.AscendingOrder):
...
-                [], QAbstractTableModel.VerticalSortHint
-            )
-        else:
-            self.layoutAboutToBeChanged.emit()
+        indices = self._sort(column, order)
+        self.__sortColumn = -1 if column < 0 else column
+        self.__sortOrder = order
+        self.setSortIndices(indices)
+
+    def setSortIndices(self, indices):
```
pavlin-policar, line 191: From what I can tell, the only reason for exposing this is that we need to be able to sync up the sorting between two table views (like in https://github.com/biolab/orange3/pull/5509), right?

orangewidget/utils/itemmodels.py
```diff
@@ -161,7 +161,10 @@ def _argsortData(self, data: numpy.ndarray, order):
         Return indices of sorted data. May be reimplemented to handle
         sorting in a certain way, e.g. to sort NaN values last.
         """
-        indices = numpy.argsort(data, kind="mergesort")
+        if data.ndim == 1:
+            indices = numpy.argsort(data, kind="mergesort")
+        else:
+            indices = numpy.lexsort(data.T[::-1])
```
pavlin-policar, line 167: I know this is fine, because I did the same thing in feature statistics, but I don't see where this would be used.

pavlin-policar (approved):
A welcome change. Looks good to me. Please merge.

orangewidget/utils/itemmodels.py
janezd: Um, yes. :) It simplifies OWPredictions a lot. But it also allows to use this model as "reordering proxy", where the order is not necessarily imposed by sorting. Is this useful in general? I don't know, maybe. But it doesn't hurt to have this in a separate method.

orangewidget/utils/itemmodels.py
janezd: `OWPredictions` uses it to sort by discrete probabilities. Both changes in this commit come from what I need there. I could derive a new class from this one (or make a new class from scratch), but I thought these changes are general enough and could also be useful elsewhere.

PR #164, opened 2021-07-20
[FIX] read_defaults: ignore schema-only settings

markotoplak:
##### Issue

The widget defaults should contain no schema-only settings. When defaults are written, schema-only settings are removed.

But saved defaults could still contain schema-only settings in some cases, such as (1) migration creates a schema-only setting (as we sometimes do for backward compatibility), or (2) we change a setting to schema-only.

This bug was found by @stuart-cls when testing Quasars/orange-spectroscopy#561. We also have a similar bug in Orange master; it was introduced In biolab/orange3#4810.

##### Description of changes
This PR removes schema-only settings when defaults are initialized, which corrects a rare bug where schema-only settings are set to a non-default value for new widgets.

Now `.provider` needs to be set on a `SettingHandler` to open defaults; it is needed to identify schema-only settings. Is this a problem? I hope not - I needed to modify some tests in widget-base, but the other stuff seems to work.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/settings.py
+++ b/orangewidget/settings.py
@@ -517,6 +517,8 @@ class SettingsHandler:
             if not isinstance(value, Setting)
         }
         self._migrate_settings(self.defaults)
+        # remove schema_only settings introduced by the migration
+        self._remove_schema_only(self.defaults)
 
     def write_defaults(self):
         """Write (global) defaults for this widget class to a file.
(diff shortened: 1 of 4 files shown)


janezd (approved):
> Now .provider needs to be set on a SettingHandler to open defaults; it is needed to identify schema-only settings. Is this a problem?

I think this is OK. I had to fix exactly the same line in #62 (which was merged but then reverted).

PR #168, opened 2021-09-07
QWebEngineView: increase send_object timeout

PrimozGodec:
##### Issue

On slow machines (sometimes on any computer) connection to receive objects can take longer than default timeout 5s. 
Reproduce: Open any widget which uses highchart (line chart in timeseries).

##### Description of changes
Increasing timeout for connection to receive objects

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/utils/webview.py
+++ b/orangewidget/utils/webview.py
@@ -533,7 +533,7 @@ elif HAVE_WEBENGINE:
             id = next(self._id_gen)
             value = self._objects[id] = dict(id=id, name=name, obj=obj)
             # Wait till JS is connected to receive objects
-            wait(until=lambda: self.receivers(self.objectChanged))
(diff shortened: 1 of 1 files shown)


janezd:
Just asking: the prospect of getting rid of highcharts in timeseries (which is, I believe, the last official add-on that requires it) is bad because some visualizations would need to be implemented from scratch in qt, right?

PrimozGodec:
> Just asking: the prospect of getting rid of highcharts in timeseries (which is, I believe, the last official add-on that requires it) is bad because some visualizations would need to be implemented from scratch in qt, right?

I made this change just as a temporary solution, it is not a good solution but will help widgets to work until we replace highcharts. I would like to get rid of highcharts in time-series but I personally have currently too many items on my to-do list.

PR #175, opened 2021-10-08
Revert "Implement signals for Setting changes"

ales-erjavec:
Reverts biolab/orange-widget-base#151

Causes 100% main thread utilization after using Silhouette Plot widget in Orange


janezd:
`_on_attr_change` is a method name that can easily appear in widgets, as a handler for changed attribute in some combo. It caused problems in Educational add-on. `__on_attr_change` would be safer.

PR #198, opened 2022-04-27
Transient preview of input and output signals

janezd:
##### Issue

Provide previewers for input and output signals.

##### Description of changes

- Refactor `MessageWidget` into a widget with a single message and a derived widget that composes a message from multiple messages (the current functionality)
- Derive a new class for messages showing input and output
- Add previewers for signals: `PartialSummary` has a new attribute for a callback that provides a preview widget

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/utils/messagewidget.py
+++ b/orangewidget/utils/messagewidget.py
@@ -103,6 +103,9 @@ class Message(
         return super().__new__(cls, Severity(severity), QIcon(icon), text,
                                informativeText, detailedText, textFormat)
 
+    def __bool__(self):
+        return not self.isEmpty()
+
     def asHtml(self, includeShortText=True):
         # type: () -> str
         """
@@ -338,13 +341,11 @@ class ElidingLabel(QLabel):
             super().setText(self.__originalText)
 
 
-class MessagesWidget(QWidget):
+class MessageWidget(QWidget):
     """
-    An iconified multiple message display area.
+    An iconified message display area.
 
-    `MessagesWidget` displays a short message along with an icon. If there
-    are multiple messages they are summarized. The user can click on the
-    widget to display the full message text in a popup view.
+    `IconifiedMessage` displays a short message along with an icon.
     """
     #: Signal emitted when an embedded html link is clicked
     #: (if `openExternalLinks` is `False`).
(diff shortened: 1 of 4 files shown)


janezd:
@ales-erjavec, could you verify that this is OK? At your convenience. For testing, you also need https://github.com/biolab/orange3/pull/5956.

Preview (implemented as popup) is closed by clicking outside it or pressing escape. Would it be complicated to close it at any click? The problem is that `mousePressEvent` is handled by widgets inside the popup. Setting `WA_TransparentForMouseEvents` is not OK because it disables scrolling as well, while `grabMouse` looks dangerous.

ales-erjavec:
> Preview (implemented as popup) is closed by clicking outside it or pressing escape. Would it be complicated to close it at any click? The problem is that `mousePressEvent` is handled by widgets inside the popup.

What about click and drag the scollbar in a  view. Click on view to set keyboard focus and navigate using keyboard <kbd>PgDown</kbd>/<kbd>PgUp</kbd> etc?

ales-erjavec:
The error/warning/info button no longer displays contents in popup on click only in tooltip and even then not the complete message. See for instance **kNN** with Data input without a target column.

Before
Screenshot_20220506_151530


Now
Screenshot_20220506_152200

janezd:
@ales-erjavec, I believe I fixed the bug with tooltips and clicks.

I guess I now also understand your earlier comment -- some functionality requires accepting mouse clicks, so we should keep it as it is? I agree; closing on Esc and with clicks outside the popup is intuitive enough.

orangewidget/utils/messagewidget.py
```diff
@@ -547,54 +630,20 @@ def flashIcon(self):
...
-        self.anim.start(QPropertyAnimation.KeepWhenStopped)
-
-        self.layout().activate()
+        super().setMessage(self.summarize())
 
     def mousePressEvent(self, event):
         if event.button() == Qt.LeftButton:
-            if self.__fulltext:
+            message = self.message
+            if message and (message.detailedText or message.informativeText):
```
ales-erjavec, line 639:
```suggestion
            if message:
```
The popup should always display if there are any contents. The inline display can frequently be too short to display the full contents.

janezd: I thought a tooltip would suffice, but I agree there's not need to disable the popup. I changed the condition as suggested.

PR #204, opened 2022-06-13
GuiTest: Undo pyqtgraph's excepthook override

ales-erjavec:
##### Issue

pyqtgraph installs a `sys.excepthook` to suppress errors in slots/event handlers. This can mask errors in tests which happen in Qt invoked slots/event handlers, ...

Reset the `sys.excepthook` to let the test abort on error. 

##### Description of changes

Undo pyqtgraph's excepthook override, let the program abort on unhandled exception in a slot, event handler, ...

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/doc/requirements-rtd.txt
+++ b/doc/requirements-rtd.txt
@@ -1,7 +1,7 @@
 --only-binary PyQt5,matplotlib
 
 setuptools
-sphinx>=4.2.0
+sphinx~=4.2.0
 AnyQt
 PyQt5~=5.12.1
 PyQtWebengine~=5.12.1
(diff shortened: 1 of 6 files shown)


janezd:
@ales-erjavec, can you unpin sphinx? #206 fixes it, I guess.

While at it, one of #206's tests failed with the same random error that we've seen in the morning. Do you have any idea why?

janezd:
@ales-erjavec is there a reason for pinning sphinx version - other than what was already fixed in #206?

ales-erjavec:
> @ales-erjavec is there a reason for pinning sphinx version - other than what was already fixed in #206?

The docs/requirements-rtd.txt specifies the build env for docs on readthedocs. As such it should pin specific sphinx version so that builds are predictable/repeatable.

PR #208, opened 2022-07-17
Fix resolution when saving images and copying to clipboard on high-dpi displays

janezd:
##### Issue

Images were stored in logical resolution instead of physical.

##### Description of changes

- For pyqtgraph objects, reimplement `ImageExporter` to observe the resolution.
- For other graphics, do the same :)
- ~~Do not use QGraphicsScene's rect for resolution; this doesn't work for widgets with mapping, e.g. `OWSOM`~~.
- Widgets that transformed the scene (e.g. OWSOM, which used grid coordinates in scene) had to export the view, not the scene. With this PR, the widget can export a scene and it will be rendered by the view, which will keep transformations, but export entire scene, not just the view.

~~**To discuss**:~~
~~- Widgets that transform scenes, such as `OWSOM` must export view, not scene because scene's render ignores transformations. To export scenes (which would also allow adjusting to any png resolution, eventually), I tried constructing a temporary view of fixed size, but it refused to render anything. On the other hand: is exporting views really bad? If the user zooms in a particular part of a chart, (s)he probably wants to export that view and not the whole scene. Maybe we should always export a view.~~

##### Includes
- [X] Code changes

### DIFF
--- a/orangewidget/io.py
+++ b/orangewidget/io.py
@@ -4,10 +4,10 @@ import tempfile
 from collections import OrderedDict
 
 from AnyQt import QtGui, QtCore, QtSvg, QtWidgets
-from AnyQt.QtCore import QMarginsF, Qt
+from AnyQt.QtCore import QMarginsF, Qt, QRectF, QPointF
 from AnyQt.QtGui import QPalette
 from AnyQt.QtWidgets import (
-    QGraphicsScene, QGraphicsView, QApplication
+    QGraphicsScene, QGraphicsView, QApplication, qApp
 )
 
 from orangewidget.utils.matplotlib_export import scene_code
@@ -77,8 +77,12 @@ class ImgFormat(metaclass=_Registry):
         raise NotImplementedError
 
     @staticmethod
-    def _get_target(scene, painter, buffer, source):
+    def _get_target(source):
         return QtCore.QRectF(0, 0, source.width(), source.height())
+
+    @classmethod
+    def _setup_painter(cls, scene, painter, source_rect, buffer):
+        pass
 
     @staticmethod
     def _save_buffer(buffer, filename):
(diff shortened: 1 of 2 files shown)


janezd:
Commits are ordered by decreasing confidence. :)

The last one may be problematice. We have a matrix of combinations here.
- The item that we are saving can be "something" from pyqtgraph, a `QGraphicsScene`, `QGraphicsView` or any other `QWidget`.
- The buffer into which we are saving can be a bitmap (`QPixmap`, `QImage`) or vector graphics (SVG).

I am not sure the code (in particular the last commit) covers all of the above, and, then, it works for any combination of the above and the bellow. For instance, saving scene to vector format may require different resolutions and bounding boxes than ... whatever else.

janezd:
@ajdapretnar, @markotoplak, can you check how this works for you? (Requires qt >= 5.14.) It should affect copying visualizations (via Ctrl-C/Cmd-C) and saving them: now their resolution should match that of the screen (e.g. double on retina).

Something else that now became possible: a widget can freely choose `graph_name = "scene"` or `graph_name = "view"` (whatever their actual names). Before, exporting a scene didn't work if the scene was transformed (i.e. the scene used a simpler coordinates, like grid coordinates in SOM). The rule should be

- Views should be exported in widgets that allow the user to zoom into parts of visualization, so (s)he's able to export those parts.
- Other widgets (that is: most, if not all) must export the scene because the view will almost always have some extra white space because of constant ratio.

@markotoplak, I consider you an expert and would appreciate your opinion. :)

ajdapretnar:
It works well for me. The resolution is good, ctrl+c, ctrl+v worked well. I tried various widgets, Scatter Plot, SOM, HierClust, Tree Viewer, Mosaic. No issues.

borondics:
Trying it only for a few saves and copy-pastes it works very nice for me.
It seems that it gets the resolution of the retina screen correctly.

However, I would like if the default DPI is set to 150 instead of 72...

PR #210, opened 2022-07-19
[ENH] Indicate dirty state on the commit button

ales-erjavec:
##### Issue

Ref: https://github.com/biolab/orange-widget-base/issues/226
Peek 2022-07-19 16-23

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -23,7 +23,7 @@ from AnyQt.QtWidgets import (
 )
 
 from orangewidget.utils import getdeepattr
-from orangewidget.utils.buttons import VariableTextPushButton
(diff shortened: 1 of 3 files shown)


janezd:
Amazing! Kudos for `_FakePaintDevice` trick.

I am not a fan of changing fonts in buttons, I don't think they look nice. I also wouldn't decorate on clean.

Perhaps a sufficiently visible symbol (like a bell emoji in smaller font - below I scaled it to 80%)?

https://user-images.githubusercontent.com/2387315/180028850-0321b140-68fe-4b80-8636-6bb1fdbcd638.mov

If emojis may be a problem, your trick allows adding anything that painter can paint, right?

PR #214, opened 2022-10-07
[FIX] ListViewSearch: Filter on changed data, introduce ListViewFilter

### LINKED ISSUE #212: ListViewSearch: Does not filter on changed data
##### Orange version
orange-widget-base:            4.18.0.dev0+86a1dd8 
Orange3:                       3.32.0.dev0+2b857a3 

##### Expected behavior
I enter a text in "Filter..." and the variables are filtered.

##### Actual behavior
Filtering does not work properly if the input data has changed.
After the filter text is deleted and re-entered it works.
When the data changes again, it does not work again.

##### Steps to reproduce the behavior

- Load Iris Dataset 
- Connect Data to Widget with ListViewSearch (I tested with Unique and Group by)
- Type "sepal" --> Only iris is filtered
- remove filter text and type again "sepal" --> iris, petal length and petal width are filtered
- remove connection between the widgets and connect again
- Type "sepal" --> Only iris is filtered

The same behavior if the data is changed e.g. with Edit Domain.

VesnaT:
##### Issue

Fixes #212

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/utils/listview.py
+++ b/orangewidget/utils/listview.py
@@ -1,4 +1,5 @@
-from typing import Iterable
+from typing import Iterable, Optional
+import warnings
 
 from AnyQt.QtWidgets import QListView, QLineEdit, QStyle
 from AnyQt.QtGui import QResizeEvent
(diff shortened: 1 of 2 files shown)


orangewidget/utils/listview.py
```diff
@@ -68,6 +68,12 @@ def setModel(self, model: QAbstractItemModel) -> None:
         self.__pmodel.setSourceModel(model)
         self.__filter_reset()
         self.model().rowsInserted.connect(self.__model_rowInserted)
+        self.model().modelReset.connect(self.__on_modelReset)
+
+    def __on_modelReset(self):
+        self.__filter_reset()
+        self.__pmodel.setFilterFixedString("")
+        self.__pmodel.setFilterFixedString(self.__search.text())
```
janezd, line 76: I don't like this. The goal is, obviously, to trigger some signal or some kind of update by changing the filter. Couldn't you trigger this directly? Apparently, setting the filter calls [`filter_changed`](https://github.com/radekp/qt/blob/master/src/gui/itemviews/qsortfilterproxymodel.cpp#L965), which you can trigger by calling `self.__pmodel.invalidateFilter()`. I tried this, but 

```python
    def __on_modelReset(self):
        self.__filter_reset()
        self.__pmodel.setFilterFixedString(self.__search.text())
        self.__pmodel.invalidateFilter()
```

doesn't work. Could you explain (perhaps next Friday, if we don't meet earlier), which side effect are you triggering by setting the filter to `""` and back again?

PR #225, opened 2023-01-09
BaseWidget: Allow keywords as strings

janezd:
##### Issue

When translating widget, e.g. to Slovenian, one can translate keywords (literally, or replace the by other words), but can't increase their number. This would be desirable because some languages may have more applicable keywords, and also because we often want to keep the original, English keywords.

##### Description of changes

Keywords can now also be given as a string (separated with commas or just spaces). Meta class converts them to strings.

This assumes that keywords are always single words. I believe this to be harmless; if somebody uses a keyword "classification tree", this will simply match both, "classification" and "tree".

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/orangewidget/tests/test_widget.py
+++ b/orangewidget/tests/test_widget.py
@@ -67,6 +67,27 @@ class WidgetTestCase(WidgetTest):
         with self.assertRaises(AttributeError):
             setattr(widget, 'unknown_field2.field', 6)
 
+    def test_keywords(self):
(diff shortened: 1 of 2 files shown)


janezd:
Change delimiter to `|`.

janezd:
If the string is separated by commas, it is split by commas (and keywords are trimmed), otherwise it's split by spaces.

PR #227, opened 2023-01-20
Report: improve report_table

janezd:
##### Issue

Fixes https://github.com/biolab/orange3/issues/6209.

Screenshot 2023-01-20 at 21 22 47

##### Description of changes

- Removes border around selected cells.
- For selected cells, it ignores colors by model and uses highlight colors from palette.
- Uses delegate's `displayText` (if it exists) to format the value
- Puts decoration in the same line (`display: inline`) and vertically centers it.
- Adds an argument `indicate_selection` that can be set to `False` in widgets that do not want to show selection in reports.

##### Includes
- [X] Code changes

### DIFF
--- a/orangewidget/report/report.py
+++ b/orangewidget/report/report.py
@@ -6,9 +6,9 @@ from collections.abc import Iterable
 from typing import Optional
 
 from AnyQt.QtCore import (
-    Qt, QAbstractItemModel, QByteArray, QBuffer, QIODevice,
+    Qt, QAbstractItemModel, QByteArray, QBuffer, QIODevice, QLocale,
     QSize)
-from AnyQt.QtGui import QColor, QBrush, QIcon
+from AnyQt.QtGui import QColor, QBrush, QIcon, QPalette
 from AnyQt.QtWidgets import QGraphicsScene, QTableView, QMessageBox
 
 from orangewidget.io import PngFormat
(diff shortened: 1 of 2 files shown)


orangewidget/report/report.py
```diff
@@ -230,24 +230,50 @@ def data(role=Qt.DisplayRole,
...
                     return data_
 
+                value = data()
+                if view is not None and col is not None:
+                    delegate = view.itemDelegateForColumn(col)
+                    if delegate is None:
+                        delegate = view.itemDelegate()
+                    if hasattr(delegate, "displayText"):
+                            value = delegate.displayText(value, QLocale())
+                value = value or ""
```
noahnovsak, line 243: There's an extra indent there that bothers me a little bit :)
Also, it confuses me that the delegate would be needed here. Shouldn't the model already be returning some nicely formatted text for DisplayRole?

Digging a bit deeper I see that specifically in the case of BarRatioTableModel, it changes DisplayRole to EditRole for some reason.

noahnovsak (approved):

orangewidget/report/report.py
janezd: > There's an extra indent there that bothers me a little bit :)

What lint doing?

(Hm, I see: no lint.)

I also thought that model would suffice, but at least some view use `displayText` for this.

Issue #226, opened 2022-06-14
Indicate uncommitted changes at the widget and canvas levels


pisarik:
Hi! As always, during summer school we are getting some new ideas to improve Orange:)

**What's your use case?**
During summer school lectures, students followed some instructions in Orange and often we had to remind them to not forget to press `Commit` button after they made changes to widget's settings (most often in Preprocessing widget).

**What's your proposed solution?**
That would be nice to indicate that the settings of a widget were changed and the `Commit` button likely have to be pressed (when auto-commit is disabled). The way to indicate it might be to change a color of `Commit` button from gray to red or something like this. Additionally that would be cool to add a red badge on a widget icon at canvas when the changes of its settings were not applied.

janezd:
Closed via #228.

PR #228, opened 2023-01-24
auto_commit: Disable button on clean, enable on dirty

janezd:
##### Issue

Ref: #226.

##### Description of changes

#210 allows for changing push buttons, but with @BlazZupan and @markotoplak we couldn't agree on clear, yet aesthetically pleasing way to indicate a dirty state. I initially didn't like their alternative idea of indicating dirty/clean state by enabling/disabling the button, but now that I tried it I find it simple and clean.

I think we should keep the code from #210 (just modified to a more general form) for potential use elsewhere.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/orangewidget/gui.py
+++ b/orangewidget/gui.py
@@ -1972,7 +1972,7 @@ def auto_commit(widget, master, value, label, auto_label=None, box=False,
                 do_commit()
         else:
             btn.setText(label)
-            btn.setEnabled(True)
+            btn.setEnabled(is_dirty())
         if callback:
             callback()
 
@@ -1982,6 +1982,7 @@ def auto_commit(widget, master, value, label, auto_label=None, box=False,
 
         def set_dirty(state):
             commit.dirty = state
+            btn.setEnabled(state)
     else:
         dirty = False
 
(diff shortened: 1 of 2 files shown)


orangewidget/gui.py
```diff
@@ -2009,8 +2011,6 @@ def do_commit():
         finally:
             QApplication.restoreOverrideCursor()
 
-    set_dirty(False)
```
janezd, line 2012: This is not needed because `toggle_checkbox` calls it.

If decide to keep it, it must be moved to the same place where I moved `checkbox_toggled`.

orangewidget/gui.py
```diff
@@ -2077,6 +2076,8 @@ def do_commit():
         setattr(master, 'unconditional_' + commit_name, commit)
         setattr(master, commit_name, conditional_commit)
 
+    checkbox_toggled()
```
janezd, line 2079: This must be moved here, otherwise a widget that inherits an autocommit and does not redecorate the commit function will fall with uninformative `AttributeError` in `checkbox_toggled` before failing with `RuntimeError` that is triggered above.

PR #229, opened 2023-01-25
Use signal's class-attribute name as default id

janezd:
##### Issue

This is needed for https://github.com/biolab/orange-canvas-core/pull/257 to have an effect, but can be merged independently.

##### Description of changes

Most signals do not have ids, but if they are defined in a class, the name of the attribute that they are assigned to can be used as default id.

This PR also checks that id's and names are unique (which should had already been an error), and warn if the same name and id do not refer to different signals (which is required for backward compatibility of https://github.com/biolab/orange-canvas-core/pull/257).

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/orangewidget/widget.py
+++ b/orangewidget/widget.py
@@ -111,9 +111,9 @@ class WidgetMetaClass(type(QDialog)):
     # pylint: disable=bad-classmethod-argument
     def __new__(mcs, name, bases, namespace, openclass=False, **kwargs):
         cls = super().__new__(mcs, name, bases, namespace, **kwargs)
+        cls.convert_signals()
         if not cls.name: # not a widget
             return cls
-        cls.convert_signals()
         cls.settingsHandler = \
             SettingsHandler.create(cls, template=cls.settingsHandler)
         return cls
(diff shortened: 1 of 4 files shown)


orangewidget/widget.py
```diff
@@ -111,9 +111,9 @@ class WidgetMetaClass(type(QDialog)):
     # pylint: disable=bad-classmethod-argument
     def __new__(mcs, name, bases, namespace, openclass=False, **kwargs):
         cls = super().__new__(mcs, name, bases, namespace, **kwargs)
+        cls.convert_signals()
```
janezd, line 114: `convert_signals` is now also called

janezd, line 114: `convert_signals` is now also called for base classes for widgets, which are themselves not widgets.

This makes assignment of id's cleaner because it does not use MRO and checks each class just once.

A side-effect is that base class must define handlers for input signals; before, it suffice that derived class (an actual widget) did so. If this is a problem, I'd put assignment of ids out of `convert_signals` and call it separately.

PR #231, opened 2023-02-10
[ENH] Add SeparatorItem and LabelledSeparator

janezd:
##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/utils/itemmodels.py
+++ b/orangewidget/utils/itemmodels.py
@@ -15,8 +15,10 @@ from AnyQt.QtCore import (
 )
 from AnyQt.QtCore import pyqtSignal as Signal
 from AnyQt.QtWidgets import (
-    QWidget, QBoxLayout, QToolButton, QAbstractButton, QAction
+    QWidget, QBoxLayout, QToolButton, QAbstractButton, QAction,
+    QStyledItemDelegate
 )
+from AnyQt.QtGui import QPalette, QPen
 
 import numpy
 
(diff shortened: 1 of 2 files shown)


orangewidget/utils/itemmodels.py
```diff
@@ -496,11 +498,48 @@ def remove(self, val):
...
+            return
+
+        painter.save()
+        palette = option.palette  # type: QPalette
+        rect = option.rect  # type: QRect
+        if data.label:
+            y = int(rect.bottom() - 0.1 * rect.height())
+            brush = palette.brush(QPalette.Active, QPalette.WindowText)
+            font = painter.font()
+            font.setPointSize(font.pointSize() * 0.9)
```
markotoplak, line 525: Crashes for me (Python 3.10, Qt 6), because `setPointSize` expects an int. Use `setPointSizeF` (and `pointSizeF`).

This also means that this delegate's paint is never tested.

janezd: Thanks. I fixed that and added a test. It doesn't tests arguments of all calls to painter, but at least test all paths.

markotoplak:
@janezd, thanks!

markotoplak:
Hm, merging this caused some tests failures. I need to check whether those tests are badly written or it is an actual bug.

borondics:
Trying it only for a few saves and copy-pastes it works very nice for me.
It seems that it gets the resolution of the retina screen correctly. 

However, I would like if the default DPI is set to 150 instead of 72...

PR #241, opened 2023-04-04
Fix image export

janezd:
##### Issue

#208 introduced some issues.

##### Description of changes

- Fixes https://github.com/biolab/orange3/issues/6391 by casting dimensions that are passed to `QImage` to ints. (`QPaintDevice` has methods `devicePixelRatio` and `devicePixelRatioF`, where the latter is defined as `qreal devicePixelRatioF()  const { return devicePixelRatio(); }`. Thanks, Qt, for misleading.)
- Fixes https://github.com/biolab/orange3/issues/6392 by handling instances of `QGraphicsWidget` and `QGraphicsView` instead of just pyqtgraph's `GraphicsWidget` and `GraphicsView`.
- Fixes rendering of `QWidget`s; `render` was called with `QPointF` instead of with `QPoint`. This disabled saving of content of Image Viewer (which doesn't work well anyway because it clips the content).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/io.py
+++ b/orangewidget/io.py
@@ -2,12 +2,13 @@ import os
 import sys
 import tempfile
 from collections import OrderedDict
+from typing import Union
 
 from AnyQt import QtGui, QtCore, QtSvg, QtWidgets
-from AnyQt.QtCore import QMarginsF, Qt, QRectF, QPoint, QRect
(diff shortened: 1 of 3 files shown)


janezd:
The first three commits contain simple fixes and need to be merged before release.

The last commit adds type-checking, which can break something. This commit can be moved into another PR to be merged later.

PR #242, opened 2023-04-04
WidgetTest: Add a test for image export

janezd:
##### Issue

See the comment in https://github.com/biolab/orange3/issues/6391. Before merging, we must see that this test works on the headless CI server.

##### Includes
- [X] Tests

### DIFF
--- a/orangewidget/tests/base.py
+++ b/orangewidget/tests/base.py
@@ -241,6 +241,12 @@ class WidgetTest(GuiTest):
                 self.skipTest("minimum size not tested as .widget was not set")
             self.check_minimum_size(widget)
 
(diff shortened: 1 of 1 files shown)


orangewidget/tests/base.py
```diff
@@ -241,6 +241,12 @@ def test_minimum_size(self):
                 self.skipTest("minimum size not tested as .widget was not set")
             self.check_minimum_size(widget)
 
+        def test_image_export(self):
+            widget = getattr(self, "widget", None)
+            if widget is None:
+                self.skipTest("minimum size not tested as .widget was not set")
```
markotoplak, line 247: The message is wrong. :)

PR #246, opened 2023-04-21
Add dpi info to saved PNGs, and scale images in reports

### LINKED ISSUE #240: Images in reports are too large
After merging #208, images in reports are too large.

To reproduce, try reporting a box plot.

I assume that WebKit treats all images as 72dpi. Perhaps adding `zoom: 0.5;` to css would fix it. **Edit:** yes, but only if the image was taken on high-dpi screen.

janezd:
##### Issue

Fixes #240.

##### Description of changes

- `report_plot` scales the image (via CSS' `zoom` property) if the image writer provides a dict with meta data that includes pixel ratio.
- DPI information is now also included in saved .png files.
- ... and in clipboard. (This seems to have no effect on macOS, but the problem is in Qt. Even if I do `image = QImage("x.png"); QApplication.clipboard().setImage(image)`, where "x.png" is a 144 dpi png, the image in the clipboard is twice the size, with 72 dpi. It's no problem, however; the user will usually resize the pasted image to some convenient size.)

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/io.py
+++ b/orangewidget/io.py
@@ -97,6 +97,25 @@ class ImgFormat(metaclass=_Registry):
         raise NotImplementedError
 
     @staticmethod
+    def _meta_data(buffer):
+        meta_data = {}
+
+        try:
+            size = buffer.size()
+        except AttributeError:
+            pass
+        else:
+            meta_data["width"] = size.width()
+            meta_data["height"] = size.height()
+
+        try:
+            meta_data["pixel_ratio"] = buffer.devicePixelRatio()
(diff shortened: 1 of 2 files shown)


markotoplak:
With this PR I understand that we get either 72 or 144. But I saw that dpi was saved correctly before on my computer (it is set to 96 dpi), so I wonder if it was also saved correctly on your Mac on master. If it is, I'd remove the dpi-setting code from the png exporter.

PR #249, opened 2023-05-31
File dialog: Add as much path as necessary to make names unambiguous

janezd:
##### Issue

Fixes https://github.com/biolab/orange3/issues/6453.

##### Description of changes

In file dialog combos, keep as much directories as necessary to disambiguate between the paths.

The code also allows for keeping more, for instance always keeping the last directory. In case we decide we prefer it.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/orangewidget/utils/filedialogs.py
+++ b/orangewidget/utils/filedialogs.py
@@ -1,3 +1,8 @@
+from collections import Counter
+from itertools import count
+
+import re
+
 import os
 import sys
 import typing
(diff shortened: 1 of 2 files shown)


VesnaT (commented):
Some windows tests could be added.

VesnaT:
The function is not working on Windows for:
```
paths = ["C:/Documents/Newsletters/Summer2018.pdf",
         "D:/Documents/Newsletters/Summer2018.pdf"]
```
and 
```
paths = ["C:/Documents/Newsletters/Summer2018.pdf",
         "C:/Documents/Newsletters/Summer2018.pdf"]
```

PR #253, opened 2023-09-15
[FIX] Fix axes in PNG export

markotoplak:
##### Issue
Fixes biolab/orange3#6543

##### Description of changes
Display size is set after painting. This keeps proper display size (intention of #246 that solved bug #240) and fixes the obnoxious axis problems.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/io.py
+++ b/orangewidget/io.py
@@ -301,6 +301,7 @@ class PngFormat(ImgFormat):
 
             # Copied verbatim from super;
             # changes are in three lines that define self.png
+            # and lines related to setDotsPerMeterX after painter.end()
             def export(self, fileName=None, toBytes=False, copy=False):
                 if fileName is None and not toBytes and not copy:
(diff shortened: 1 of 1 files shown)


markotoplak:
The bug was reproduced by @noahnovsak on his laptop (Mac, external display disconnected) and @thocevar (Windows, with external display connected). Could you please report if this PR fixes your issues?

@PrimozGodec, can you double check that this PR does not reintroduce #240?

PrimozGodec (approved):
It doesn't break changes in #240. I could also reproduce the issue on my MacBook (without this PR) and I can confirm it is fixed by this PR.

thocevar:
Fixed on Windows as well. 

There is another separate problem (not introduced by this fix). It doesn't take into account global scaling. With a global 125% scaling it exports a smaller image which has a lower quality when rescaled to match the image on the screen.

PR #256, opened 2023-10-15
[RFC] Don't call eventFilter on destroyed objects

janezd:
##### Issue

An alternative to https://github.com/biolab/orange3/pull/6596.

Qt may call `eventFilter` for objects that have already been destroyed and have an empty `__dict__`. This may cause crashes, in particular in unit tests.

##### Description of changes

This PR resolves some of those cases by wrapping `eventFilter` into a method that redirects such calls directly to the `eventFilter` inherited from `QDialog`. Note that this only works for classes derived from `OWWidget`. Most `eventFilter` methods that we use in Orange are defined in other classes.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/tests/test_widget.py
+++ b/orangewidget/tests/test_widget.py
@@ -758,5 +758,36 @@ class TestWidgetMenu(WidgetTest):
         self.assertFalse(mb.isVisibleTo(w))
 
 
+class MetaClassTest(WidgetTest):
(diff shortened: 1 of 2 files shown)


ales-erjavec:
This would not effect https://github.com/biolab/orange3/pull/6596 and is not worth messing with metaclasses for.

PR #270, opened 2024-03-25
ListViewFilter: Retain selection when filtering

VesnaT:
##### Issue

Fixes https://github.com/biolab/orange3/issues/6768

##### Description of changes
The default selection corresponds to `QSortFilterProxyModel`. 
Apply selection mapped to a source model.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/orangewidget/utils/listview.py
+++ b/orangewidget/utils/listview.py
@@ -1,3 +1,4 @@
+from contextlib import contextmanager
 from typing import Iterable, Optional
 import warnings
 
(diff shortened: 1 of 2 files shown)


orangewidget/utils/listview.py
```diff
@@ -10,9 +11,19 @@
...
 )
 
 
+@contextmanager
+def disconnected(signal, slot, connection_type=Qt.AutoConnection):
+    signal.disconnect(slot)
+    try:
+        yield
+    finally:
+        signal.connect(slot, connection_type)
```
janezd, line 24: Perhaps blocking signals (https://doc.qt.io/qt-6/qobject.html#blockSignals) would be better the temporarily disconnecting them?

Also: search the code for `def blocked(`. You'll find that a similar function already exists in two other places. Perhaps we should introduce one in `orange-widget-base` and then (after release...) use it wherever needed?

VesnaT: Or even use an existing one at 
https://github.com/biolab/orange-widget-base/blob/6ddb801d2467d2f2970316b99100fb15df2d911e/orangewidget/utils/itemmodels.py#L43

PR #275, opened 2024-06-11
Multilingual configuration

janezd:
Multilingual configuration

### DIFF
--- a/i18n/si/msgs.jaml
+++ b/i18n/si/msgs.jaml
@@ -302,9 +302,9 @@ settings.py:
             values=: false
             ): false
 version.py:
-    4.24.0: false
-    4.24.0.dev0+31e1b65: false
-    31e1b655e24581af44ab5f14523ace5a364f57a8: false
+    4.25.0: false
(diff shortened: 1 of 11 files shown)


i18n/si/msgs.jaml
```diff
@@ -302,9 +302,9 @@ settings.py:
             values=: false
             ): false
 version.py:
-    4.24.0: false
-    4.24.0.dev0+31e1b65: false
-    31e1b655e24581af44ab5f14523ace5a364f57a8: false
+    4.25.0: false
```
markotoplak, line 305: I think we need options to ignore whole parts (files, classes, methods).

i18n/trans.sh
```diff
@@ -1,8 +1,8 @@
-if [ "$#" -ne 2 ]
```
markotoplak, line 1: Do we still need this file?

markotoplak:
Almost works. Making a sdist and then installing it fails because some files are missing from the MANIFEST.in

PR #278, opened 2024-09-19
open_filename_dialog: Allow reading *.*

janezd:
##### Issue

`open_filename_dialog` has an argument `add_all` which, if `True`, adds "All readable files" that allows loading a file with any known extension. This is not enough to implement https://github.com/biolab/orange3/issues/6792.

##### Description of changes

- If `add_all` is set to `*` it offers all extensions (*.*), not just the known.
- An unrelated improvement: if there is only a single file format, "All readable files" makes no sense and is not shown. This is situation might be common - there may be only a single file format and the caller forgets to change the default value of `add_all` from `True` to `False`.
- 
- ##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation

### DIFF
--- a/orangewidget/utils/filedialogs.py
+++ b/orangewidget/utils/filedialogs.py
@@ -162,8 +162,8 @@ def open_filename_dialog(start_dir: str, start_filter: str, file_formats,
         start_dir (str): initial directory, optionally including the filename
         start_filter (str): initial filter
         file_formats (a list of FileFormat): file formats
-        add_all (bool, default True): add a filter for all supported extensions.
-           If set to `*`, show an option to read all files, *.*.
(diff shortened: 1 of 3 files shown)


orangewidget/utils/filedialogs.py
```diff
@@ -162,7 +162,8 @@ def open_filename_dialog(start_dir: str, start_filter: str, file_formats,
         start_dir (str): initial directory, optionally including the filename
         start_filter (str): initial filter
         file_formats (a list of FileFormat): file formats
-        add_all (bool): add a filter for all supported extensions
+        add_all (bool, default True): add a filter for all supported extensions.
```
markotoplak, line 165: Not just bool anymore...

PR #282, opened 2024-12-13
[FIX] widget: Make controlAreaVisible a global setting

ales-erjavec:
##### Issue

Make controlAreaVisible a global setting along the widget geometry.

##### Description of changes

Make controlAreaVisible a global setting

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


markotoplak:
@ales-erjavec, thanks, this change means that now a newly added widget could be shown with a hidden control area. @janezd (and me) pointed out that for new widgets, the control area should always be shown. The size is problematic: if a default was saved with a hidden control area, the new widget size should be increased in width by the control area size.

PR #295, opened 2026-06-02
[FIX] Broken LabelledSeparator 

### LINKED ISSUE #294: Broken LabelledSeparator
#293 broke `LabelledSeparator`. No tests in orange-widget-base covered this, but orange3 tests fail with:

```
  File "/Users/janez/orange3/Orange/widgets/data/tests/test_owcontinuize.py", line 823, in test_contmodel
    self.assertIsNone(ind.data())
AssertionError: (None, '', False) is not None
```

`LabelledSeparator` is derived from `SeparatorItem` and adds a label. We want it to be a `Separator` to inherit other behaviour, but it contains additional data that is lost because `PyListModel.data` no longer returns after #293.

The easiest fix would be to replace `if isinstance(item, SeparatorItem): return None` with `if type(item) is SeparatorItem: return None`.

@ales-erjavec, would this be OK or should we do something more elaborate?

ales-erjavec:
##### Issue

Fixes #294

##### Description of changes

* Revert #293
* Give SeparatorItem an empty str
* Add test for SeparatorItem display text

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/.github/workflows/run-tests-workflow.yml
+++ b/.github/workflows/run-tests-workflow.yml
@@ -44,15 +44,15 @@ jobs:
 
           - os: ubuntu-latest
             python-version: "3.12"
-            test-env: "PyQt6~=6.5.0 PyQt6-Qt6~=6.5.0 PyQt6-WebEngine~=6.5.0 PyQt6-WebEngine-Qt6~=6.5.0"
-
-          - os: ubuntu-latest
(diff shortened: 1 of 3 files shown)


janezd:
So much cleaner. Thanks!

